PR #32, opened 2026-04-22
agent: token-conscious screenshots (resize + iterative JPEG compress)

esokullu:
## Summary

Ports the two image-sizing ideas from Anthropic's [Claude-for-Chrome](https://webbrain.one/blog/vision-model-shootout) extension into our screenshot pipeline, so every screenshot is cheap enough to throw at any vision endpoint — local or paid — without cost-per-image anxiety.

Two helpers on `Agent` (Chrome + Firefox parity):

- **`_fitImageDimensions(w, h, budget)`** — binary-search on the long side for the largest (w, h) that fits both a per-dimension pixel cap and a total-token cap (default 1568 / 1568 at 28 px per token). Preserves aspect ratio. Port of their `C(w, h, params)` with clearer names.
- **`_shrinkImageForBudget(dataUrl, origW, origH)`** — end-to-end pass: decode → draw at target dims → iteratively drop JPEG quality (0.75 → 0.10 in 0.05 steps) until base64 fits under `maxBase64Chars` (~1.4 MB). Best-effort, falls back to original on any failure.
- **`_compressJpegToByteCeiling(dataUrl)`** — dim-preserving variant for coord-aligned captures that can't be resized.

## Wired into

1. **`_captureAutoScreenshot`** — probes CSS viewport, asks CDP to capture at a pre-scaled resolution via `clip.scale` (so we never materialize a multi-MB native-DPR JPEG just to throw away). Byte-ceiling fallback runs afterwards as a safety net. `coord_aligned` mode skips the resize — fidelity > tokens for pixel clicks.
2. **`screenshot` tool** — same treatment. JPEG + budget-aware clip in the default path; PNG + `scale:1` in coord_aligned mode. Description string reports resized dimensions so the planner knows what it's looking at.
3. **`full_page_screenshot`** — was returning raw base64 via `{image: …}` which hit `_limitToolResult`'s 8 KB chop and mangled the payload, AND had no size ceiling. Now shrinks via the budget pass (using the decoded bitmap's dims since we don't know doc height up front) and ships through `_attachImage`.

Firefox side: MV2 background page has a real DOM so it uses regular `<canvas>` + `toBlob` instead of `OffscreenCanvas`. `captureVisibleTab` doesn't accept `clip/scale` so we always decode + resize post-capture (marginally slower than Chrome's CDP pre-scale but identical output).

## Bench (from the quick binary-search sanity check)

| Source viewport | After fit | Tokens before → after |
|---|---|---|
| 1280×800 | 1280×800 | 1307 → 1307 (no-op) |
| 1920×1080 | 1478×831 | 2645 → 1567 |
| 2560×1440 | 1478×831 | 4703 → 1567 |
| 3840×2160 | 1478×831 | 10580 → 1567 |
| 1920×8000 (full-page) | 376×1568 | **19592 → 752** |

26× cost reduction on a pathological full-page capture, always within the provider's image cap.

## Intentionally left alone

- **`done` verification screenshot.** It annotates a red outline around the last ax-interacted element and fires once per task — annotation legibility matters more than per-image cost.

## Test plan

- [ ] Load extension, run `auto_screenshot` path on a 1920×1080 viewport, confirm captured image is 1478×831 JPEG and under ~1 MB base64.
- [ ] Call `screenshot` tool explicitly with `coord_aligned: true`, confirm image is CSS-pixel-aligned PNG (no resize).
- [ ] Call `screenshot` tool with default args, confirm resize happens and description string reports the new dims.
- [ ] Call `full_page_screenshot` on a long-scroll page, confirm it no longer truncates and the image reaches the planner.
- [ ] Sanity-check Firefox parity: load Firefox MV2 build, confirm auto-screenshot + screenshot tool both produce budget-fit JPEGs.
- [ ] Run `test/vision-probe.mjs` against a post-change screenshot to confirm vision model output quality is still acceptable on a downscaled image.

### DIFF
--- a/src/chrome/src/agent/agent.js
+++ b/src/chrome/src/agent/agent.js
@@ -1023,31 +1023,55 @@ Rules: no prose intro, no conclusion, no "this screenshot shows...", no layout d
       await cdpClient.attach(tabId);
       await cdpClient.sendCommand(tabId, 'Page.enable');
       await this._bringToFrontForCapture(tabId);
+
+      // Probe the CSS viewport first so we can either (a) clip exactly
+      // to it for pixel-accurate captures, or (b) compute a budget-aware
+      // CDP-side scale that downsizes during capture rather than after.
+      const vp = await cdpClient.evaluate(tabId, '({w: window.innerWidth, h: window.innerHeight})');
+      const cssW = Math.max(1, Math.round(vp?.result?.value?.w || 1024));
+      const cssH = Math.max(1, Math.round(vp?.result?.value?.h || 768));
+
       if (coordAligned) {
-        const vp = await cdpClient.evaluate(tabId, '({w: window.innerWidth, h: window.innerHeight})');
-        const w = Math.max(1, Math.round(vp?.result?.value?.w || 1024));
-        const h = Math.max(1, Math.round(vp?.result?.value?.h || 768));
+        // Pixel-accuracy mode: image pixels must equal CSS pixels so the
+        // planner can click by coordinate off the screenshot. Skip the
+        // token-budget resize — the whole point of this mode is fidelity.
+        // We DO still run the byte-ceiling fallback afterwards: if the
+        // CSS viewport happens to be huge, we'd rather lose some JPEG
+        // quality than overflow the provider's image cap.
         const shot = await cdpClient.sendCommand(tabId, 'Page.captureScreenshot', {
           format: 'jpeg',
           quality: 60,
-          clip: { x: 0, y: 0, width: w, height: h, scale: 1 },
+          clip: { x: 0, y: 0, width: cssW, height: cssH, scale: 1 },
         });
         if (!shot?.data) return null;
-        return {
-          dataUrl: `data:image/jpeg;base64,${shot.data}`,
-          width: w,
-          height: h,
-          coordAligned: true,
-        };
+        const rawDataUrl = `data:image/jpeg;base64,${shot.data}`;
+        const shrunk = await this._compressJpegToByteCeiling(rawDataUrl);
+        return { dataUrl: shrunk, width: cssW, height: cssH, coordAligned: true };
       }
+
+      // Non-coord-aligned mode: pre-compute target dims via the budget
+      // binary-search, then ask CDP to capture + scale in one pass. This
+      // avoids ever materializing a multi-MB native-DPR JPEG that we'd
+      // then have to decode and resize in the service worker.
+      const [targetW, targetH] = Agent._fitImageDimensions(cssW, cssH);
+      const scale = targetW < cssW ? targetW / cssW : 1;
       const shot = await cdpClient.sendCommand(tabId, 'Page.captureScreenshot', {
         format: 'jpeg',
-        quality: 60,
+        quality: Math.round(Agent.IMAGE_BUDGET.initialJpegQuality * 100),
         fromSurface: true,
+        clip: { x: 0, y: 0, width: cssW, height: cssH, scale },
       });
       if (!shot?.data) return null;
+      const rawDataUrl = `data:image/jpeg;base64,${shot.data}`;
+
+      // CDP-side resize + JPEG q=75 usually fits. Iterative quality
+      // downgrade is the safety net for high-DPR screens where the
+      // captured image can still exceed the base64 ceiling.
+      const shrunk = await this._compressJpegToByteCeiling(rawDataUrl);
       return {
-        dataUrl: `data:image/jpeg;base64,${shot.data}`,
+        dataUrl: shrunk,
+        width: targetW,
+        height: targetH,
         coordAligned: false,
       };
     } catch (e) {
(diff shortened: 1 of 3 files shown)


esokullu:
Good catch on the `full_page_screenshot` behavior change. Verified it's safe — and pushed unit tests for the budget math.

**On the non-vision `full_page_screenshot` behavior change**

I grepped the full codebase for every consumer of the tool's return value and confirmed nothing internal reads the `.image` field:

- [`src/chrome/src/agent/agent.js:1946`](../blob/token-conscious/src/chrome/src/agent/agent.js#L1946) — `_summarizeToolResult` for the step-by-step UI reads `parsed.success` only, never `parsed.image`.
- The only other consumer is `_limitToolResult` → stringified tool-message content handed to the LLM. No hardcoded field access.

So the change is: a non-vision provider that calls `full_page_screenshot` now gets `{success: false, error: \"This model cannot see images…\"}` instead of a `{success: true, image: <truncated base64 garbage>}`. That's strictly better — before this PR that path was already broken (`_limitToolResult` was chopping the base64 at 8KB mid-stream, producing invalid data URLs that no decoder could handle). The model just couldn't tell it was getting garbage. Now it gets a clear refusal and falls back to `get_accessibility_tree` or `read_page`, which is the right behavior.

Worth a manual sanity check in the loaded extension anyway — will do that before merging.

**Tests added** ([`ce9ed08`](../blob/token-conscious/test/run.js))

9 new tests under `image budget` covering:

- small viewport pass-through (fast path)
- 1080p shrinks to the token cap with aspect preserved
- 4K and 1440p converge to identical max-budget dims (catches off-by-one in the binary search)
- tall portrait (1920×8000 full-page) caps the long side
- 5K retina stays within budget
- square input returns square out
- pathological 1×100000 strip terminates
- custom budget params honored
- monotonicity fuzz across common viewport sizes

All 41 tests pass via `npm test`. No browser needed — it's pure math, mirroring `Agent._fitImageDimensions` / `_estimateImageTokens` as shims the same way `LoopDetectorShim` mirrors the loop-detection helpers.

**What the unit tests intentionally don't cover**

Three things need a loaded extension:

1. **Coord-aligned captures** — the `scale: 1` clip path + actual click coordinate reliability. That's a CDP integration, not math.
2. **Live full-page captures with/without vision provider** — to verify the vision-sub-call + image_attach routing works end-to-end.
3. **Real high-DPR monitors** — my 1920×8000 test input is synthetic; what actually matters is how a retina display's real CDP capture flows through. Behavior should be identical, but CDP's quirks bite in places benchmarks miss.

Plan: load the build, run the `state_change` auto-screenshot loop against a long-scroll page (e.g. a GitHub issues list), and confirm the captured image is JPEG-encoded and under ~1 MB base64 per step. If that holds, merge.

esokullu:
## Manual validation plan before merge

Six tests. Run them in order — tests 1–2 validate the "happy path" resize is working, tests 3–5 validate the three `full_page_screenshot` provider-routing branches, test 6 is the high-DPR confirmation. Each test lists what to do, what to expect, and how to inspect.

### One-time setup

1. Build the unpacked extension and load it in Chrome:
   - `chrome://extensions` → Developer mode on → **Load unpacked** → select `src/chrome/`.
2. Open the service worker DevTools so you can see captures as they happen:
   - On the same `chrome://extensions` row, click the **service worker** link.
   - Keep the **Console** tab open.
3. In WebBrain settings:
   - **Display tab** → Auto-screenshot: `On state change (default)`
   - **Display tab** → Verbose Mode: ON (so tool results show raw JSON in the side panel)
   - **Providers tab** → pick your primary (e.g. llama.cpp at your endpoint). Leave `Model supports vision` OFF for now; we'll toggle it per test.
   - **Vision tab** → leave blank for now.
4. Have one long-scroll test page bookmarked. Recommended: a Wikipedia article with a table of contents (e.g. `https://en.wikipedia.org/wiki/Attention_(machine_learning)`) or the WebBrain repo's **Issues** page on GitHub.

### How to check a screenshot's size without any instrumentation

After any run, open the side-panel and note the tool result card for the screenshot-ish tool. In verbose mode it shows the result JSON including the description field, which now includes dimensions and byte count. For deeper inspection, the verbose-mode button accepts **shift+click** to dump the full LLM request/response log — the `image_url` blocks will be in there. Copy a dataUrl, paste into DevTools console as `len = "data:image/jpeg;base64,...".length`, and confirm `len < 1_398_100`.

### Test matrix

| # | What | Settings | Send in side panel | Expect |
|---|---|---|---|---|
| 1 | Auto-screenshot viewport resize | Main vision: **ON** | "scroll down three times" on a Wikipedia article | 3 auto_screenshot events. For each, the image sent to the model is `data:image/jpeg;base64,...` with length < 1.4M chars. On a 1080p display: dimensions ≈ 1478×831. On 1440p/4K: same 1478×831 (converges). |
| 2 | coord_aligned pixel accuracy | Main vision: **ON** | Act mode: "take a screenshot with coord_aligned set to true" then "click at x=50 y=200" | Tool description says "CSS-pixel aligned for pixel clicks"; image is **PNG** (not JPEG); image dimensions exactly match the CSS viewport (no resize applied). Follow-up click lands on the element under that CSS coordinate — verify by eye. |
| 3 | full_page with dedicated vision provider | Main vision: off. **Vision tab: configured** (e.g. pointing at llama.cpp + Qwen3.6-35B-A3B). | Act mode: "take a full page screenshot" | Tool result `{success: true, method: "vision_describe", description: "[Full-page screenshot described by vision model ...]\n..."}`. **No raw image** in the result. Service worker console shows a `vision_sub_call` trace event. |
| 4 | full_page with only main-provider vision | **Main vision: ON**. Vision tab: blank. | Act mode: "take a full page screenshot" | Tool result `{success: true, method: "image_attach", description: "Full page screenshot captured and fit to vision budget (WxH, N base64 chars)"}`. An `image_url` block is attached to a follow-up user message. Image is JPEG, W×H respects budget (long side ≤ 1568, token cost ≤ 1568), base64 length < 1.4M. |
| 5 | full_page with no vision anywhere (behavior-change test) | **Main vision: OFF**. Vision tab: blank. | Act mode: "take a full page screenshot" | Tool result `{success: false, error: "This model cannot see images..."}`. The agent should gracefully switch strategies (e.g. call `read_page` / `get_accessibility_tree`) rather than retry. This is the reviewer's concern — verify the planner is not confused by the refusal. |
| 6 | High-DPR / retina behavior | Main vision: **ON**. If you don't have a retina display, simulate via DevTools → Device Toolbar → set `device pixel ratio = 2` on the active tab. | "take a screenshot" (default, non-coord) | JPEG output; dimensions in the description should be ≈ `_fitImageDimensions(cssW, cssH)` — NOT the native-DPR pixel count. On a simulated 2×-DPR 1280×800 tab, you should still get ~1280×800 *capped* at the budget target (≈1280×800 unchanged since it's under the 1568 cap), NOT 2560×1600. This confirms `clip.scale` is doing its job and we're not materializing a native-DPR image just to throw it away. |

### Cross-cutting invariants (check on every test)

- [ ] Every screenshot reaching the model is **JPEG**, except test 2 which must be **PNG** (coord-aligned) because fidelity matters there.
- [ ] Every base64 length is under **1,398,100 chars** (~1.4 MB). If any crosses it, the iterative JPEG quality loop should have kicked in — add a one-line `console.log('[screenshot] quality=', quality)` inside the while-loop in `_shrinkImageForBudget` temporarily if you want to see which step it exited at.
- [ ] Auto-screenshot after scroll/click/type all produce the same budget-fit dims (no regression in the CDP clip path).
- [ ] Service worker memory stays flat over 20+ consecutive auto-screenshots (no leak from repeated `createImageBitmap` / `OffscreenCanvas` without release — Chrome usually GCs them, but worth watching in DevTools → Memory tab).

### If any test fails

Post the details in a comment with:
- Which test (#)
- The full tool-result JSON from the side panel
- The `description` field value (has dims + bytes)
- Whatever's relevant from the service worker console (errors, warnings)

### Merge criteria

All 6 tests pass their "Expect" column **and** the cross-cutting invariants hold. Test 5 in particular has to pass — that's the behavior change the reviewer flagged. If the planner gets confused by the `{success: false, error: "..."}` response on the no-vision path and loops, we need to soften the refusal or add a hint suggesting `read_page` / `get_accessibility_tree` explicitly.

Roughly 30 minutes to run end-to-end, assuming a working llama.cpp endpoint and a display the tests can hit.

PR #46, opened 2026-05-04
Claude (Pro/Max subscription) provider via OAuth

esokullu:
## Summary

Adds a new **Claude (Pro/Max subscription)** provider entry that authenticates with the user's Claude.ai account via OAuth instead of an API key. The existing API-key Anthropic provider stays untouched — users can have both configured side by side and switch between them.

OAuth re-uses the flow Claude Code (Anthropic's own CLI) ships with, including its public `client_id`. Tokens land in `chrome.storage.local` / `browser.storage.local` under `anthropicOauthTokens`, separate from provider configs.

### Implementation highlights

- `oauth-claude.js` — PKCE S256 flow. Opens auth URL in a new tab and listens via `tabs.onUpdated` for the redirect (URL change fires before console.anthropic.com finishes loading). State validation, refresh-token rotation, hard-fail wipe on revoked refresh.
- `AnthropicOAuthProvider extends AnthropicProvider` — overrides `_headers` (Bearer + `anthropic-beta: oauth-2025-04-20`, drops `x-api-key`), `_convertMessages` (auto-prefixes the mandatory `"You are Claude Code, Anthropic's official CLI for Claude."` preamble), and wraps `chat`/`chatStream` with token-refresh + 401-retry.
- Settings UI: special-cased card with sign-in button, live auth-state badge, sign-out, and a yellow disclaimer panel about ToS / reliability. Save/Test/Activate buttons share handlers with the normal provider cards.
- Mirrored to Firefox with `browser.*` APIs.

### Why no `chrome.identity.launchWebAuthFlow`?

That API requires the redirect URI to be `https://<extension-id>.chromiumapp.org/`, but Claude Code's client_id is registered with `https://console.anthropic.com/oauth/code/callback` and that's the only redirect Anthropic's auth server will accept. We open a tab and intercept the redirect via `tabs.onUpdated` instead.

### Why no `chrome.alarms` refresh?

The provider does lazy refresh on every chat call (expiry check + 401-retry safety net), so a proactive alarm would just add an `alarms` permission and a re-permission prompt at update for no real benefit.

### ToS / reliability caveats

Anthropic restricts using a Pro/Max subscription with non-Anthropic tools, and can revoke their CLI's OAuth client at any time. The disclaimer surfaces in three places: the in-card panel, README's Known Issues section, and the `oauth-claude.js` file-header comment. The Claude Code system-prompt preamble is also mandatory — Anthropic's OAuth gate flags requests that omit it; do **not** strip it.

OpenAI Codex subscription provider was scoped out of this PR (different API surface, larger lift) but the OAuth helper generalizes cleanly for a follow-up.

## Test plan

- [ ] Load the extension, open Settings → "Claude (Pro/Max subscription)" card. Status loads as "Not signed in", Sign-in button visible.
- [ ] Click Sign in with Claude → new tab opens at claude.ai/oauth/authorize. Complete authorization. Tab closes automatically. Card updates to "Signed in. Token issued <time>, refreshes around <time>". Sign-in button hidden, Sign-out shown.
- [ ] Click Set Active → side panel uses this provider on the next message. Send a chat. Verify it works (text + tool-use round trip).
- [ ] Force token expiry by setting `anthropicOauthTokens.expiresAt` to `0` in storage. Send another chat. Verify it refreshes and succeeds without user intervention.
- [ ] Click Sign out → status returns to "Not signed in". Send a chat → fails with "Not signed in to Claude" message.
- [ ] Cancel sign-in mid-flow (close the auth tab without authorizing). Verify status updates to "Sign-in failed: Sign-in window was closed before authorization completed."
- [ ] Repeat in Firefox.
- [ ] Verify the API-key Anthropic provider still works untouched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ Open-source AI browser agent for Chrome and Firefox. Chat with any web page, aut
   - **OpenAI** (GPT-5.3, etc.)
   - **OpenRouter** (access 100+ models)
   - **Anthropic Claude** (native API)
+  - **Claude (Pro/Max subscription)** — sign in with your Claude.ai account via OAuth instead of an API key. See *Known Issues* below for the ToS / reliability caveats.
 - **Side Panel UI** — Clean chat interface that lives alongside your browsing
 - **Per-Tab Conversations** — Each tab has its own chat history
 - **Streaming** — Real-time token streaming from all providers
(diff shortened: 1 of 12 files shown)


esokullu:
we should add openai subscription too

PR #120, opened 2026-06-01
upload_file: don't report success when the path probe was inconclusive

esokullu:
## Follow-up to #118

#118 was merged at `a314241`, just before the fix for the last codex finding landed — so `main` still carries that P2. This PR lands it.

## Problem

In `upload_file`, the pre-validation guard only rejected `{exists:true, readable:false}`. But `probeLocalFile` returns:

| probe result | meaning |
|---|---|
| `{exists:true, readable:true}` | real, readable file ✅ |
| `{exists:true, readable:false}` | phantom entry — path missing/unreadable ❌ (already rejected) |
| `{exists:false, readable:null}` | nothing attached — **probe inconclusive** |
| `null` | probe couldn't run — **inconclusive** |

When the probe was **inconclusive**, the path was never validated, yet the later `files.length === 0` branch still returned `success: true`. On async upload widgets that clear the real input on `change`, a bad/stale path would therefore report an unverified success — the exact false positive this pre-validation exists to prevent. Because the agent treats any `success: true` as completed work, a GitHub release flow with a stale path could **publish without its asset**.

Flagged by codex on #118 ([thread](https://github.com/esokullu/webbrain/pull/118#discussion_r3333461770)).

## Fix

Derive `pathConfirmed = !!(probe && probe.exists && probe.readable === true)` and only let the empty-`FileList` branch (and the unreadable-list fallback) report `success: true` when it holds. An inconclusive probe now returns a non-success result that tells the model to verify against the page and re-check the path if the asset is genuinely missing — avoiding both the silent false positive and a blind retry loop.

## Testing

- `npm test` → 251 passed, 0 failed
- `node --check` on the modified file passes

### DIFF
--- a/src/chrome/src/agent/agent.js
+++ b/src/chrome/src/agent/agent.js
@@ -4748,13 +4748,23 @@ Rules: no prose intro, no conclusion, no "this screenshot shows...", no layout d
         // the target input on `change`, so reading the target back can't tell
         // "consumed a valid file" from "got a bad path". A detached isolated
         // probe input answers that authoritatively without hitting delegated
-        // upload handlers on the page. Only block on a definitive
-        // "present but unreadable" verdict. A null/unknown probe falls
-        // through so we never turn a possibly-good upload into a hard failure.
+        // upload handlers on the page. The probe returns one of:
+        //   {exists:true,  readable:true}  → path is a real, readable file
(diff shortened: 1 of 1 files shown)


esokullu:
@codex review

PR #122, opened 2026-06-04
feat(deepseek): update default model to deepseek-v4-flash with 1M context window

xianhongtao:
- Change default DeepSeek model from deepseek-chat to deepseek-v4-flash
- Add contextWindow: 1000000 (1M tokens) for DeepSeek provider
- Update settings suggestions: deepseek-v4-flash, deepseek-v4-pro
- Update all README and docs to reflect new default model
- Apply changes to both Chrome and Firefox extensions


esokullu:
thank you looks good!

PR #204, opened 2026-06-27
Ayush7614/feat/history tools

esokullu:


esokullu:
Reviewed this branch and pushed a follow-up fix: `c28c239b` (`Fix Chrome history URL normalization`).

What changed:
- Chrome now preserves query params and hash in the URL normalization used by history-navigation checks, so `go_back` / `go_forward` correctly treat query/hash-only history entries as successful navigation changes, matching the Firefox behavior.
- Expanded the normalization regression test to cover both Chrome and Firefox.

Validation:
- `npm test` passed locally: 437 unit checks, 0 failed; injection corpus 59/59 passed.

esokullu:
Follow-up on Copilot review `4585827055`: both comments are already addressed by the current PR head `c28c239b` (`Fix Chrome history URL normalization`).

Details:
- Chrome `_normalizeUrl()` now preserves `search` and `hash`, matching Firefox, so query/hash-only history entries are treated as successful back/forward navigation changes.
- The normalization regression test now runs against both `AgentCh` and `AgentFx`.

The Copilot comments are attached to the previous reviewed commit `637ea5cb`, which is why the review text still describes the pre-fix state even though the branch head has moved.

Validation rerun on current head:
- `git diff --check` passed.
- `npm test` passed locally: 437 unit checks, 0 failed; injection corpus 59/59 passed.

PR #206, opened 2026-06-27
Elijah hash7/feature/mutation api observer

esokullu:


esokullu:
Reviewed this branch and pushed a follow-up fix: `dcadc1fb` (`Restrict API shortcut hints to GET requests`).

What changed:
- The API-shortcut detector now only suggests replaying repeated GET XHRs via `fetch_url`.
- Repeated POST/PUT/DELETE-style requests are ignored by the shortcut, so the loop warning does not nudge the model into API mutations without the explicit `/allow-api` path.
- Added regression coverage for write-method requests.

Validation:
- `npm test` passed locally: 441 unit checks, 0 failed; injection corpus 59/59 passed.

esokullu:
Addressed Copilot review `4585822943` in `4c3f8d5b` (`Prevent API shortcut request reuse`).

What changed:
- `_detectApiShortcut` now consumes matched API request indexes, so one recorded request cannot satisfy multiple overlapping click windows.
- Applied the fix in both Chrome and Firefox agent copies.
- Updated the test shim and added a regression where a single GET after a click burst must not produce an API shortcut.

Validation:
- `git diff --check` passed.
- `npm test` passed locally: 442 unit checks, 0 failed; injection corpus 59/59 passed.

esokullu:
Correction to my earlier GET-only follow-up: I pushed `558c64fe` (`Allow API shortcut write methods`) to keep the API shortcut detector open to POST/PUT/PATCH/DELETE-style methods as intended.

Rationale:
- The detector should surface the repeated same URL+method request pattern.
- The existing `/allow-api` / UI-vs-API policy remains responsible for whether a mutating replay is allowed.
- The request-reuse fix from `4c3f8d5b` remains in place, so one captured request still cannot satisfy multiple overlapping click windows.

What changed:
- Removed the GET-only method filter from Chrome, Firefox, and the test shim.
- Updated the shortcut warning so it no longer says "get the next page's data" and instead points back to the UI/API mutation policy for mutating methods.
- Replaced the old write-method rejection test with a POST eligibility regression.

Validation:
- `git diff --check` passed.
- `npm test` passed locally: 442 unit checks, 0 failed; injection corpus 59/59 passed.

PR #231, opened 2026-06-30
remove inspectpage

esokullu:


esokullu:
instead just cancelled from functions by returning return; because we may need this back soon

PR #234, opened 2026-06-30
agent: add adapter for sahibinden.com

Ohualtex:
Adds a site adapter for **sahibinden.com**, Türkiye's largest classifieds site. CONTRIBUTIONS.md lists Turkish regional adapters as the #1 wanted contribution.

**Real failure modes it prevents**
- Hunting for an "Add to cart" on contact-the-seller listings — sahibinden is classifieds, not a checkout store.
- Looping into sahibinden's DataDome anti-bot wall instead of stopping and surfacing it to the user.
- Re-fetching the same sorted results URL repeatedly (the exact pattern documented in TODOs.md §4) instead of extracting from the page already loaded.

It also names the Turkish UI labels (Giriş Yap, İlan Ver, Sıralama, İl/İlçe) and the "Onay Bekliyor" pending-approval state so the agent doesn't report a listing as published prematurely.

- Mirrored to both `src/chrome/` and `src/firefox/` adapters.js.
- Adds a `getActiveAdapter` match + content test; `npm test` → 509/509 passing.

### DIFF
--- a/src/chrome/src/agent/adapters.js
+++ b/src/chrome/src/agent/adapters.js
@@ -553,6 +553,23 @@ const ADAPTERS = [
 - Pre-orders / "Made to order" listings have a longer ship date — surface that to the user before buying.`,
   },
 
+  // ─── Regional — Türkiye (TR) ──────────────────────────────────────────
+  // Regional adapters are the project's #1 wanted contribution (CONTRIBUTIONS.md);
+  // Türkiye is top of the priority list. Add more TR sites (trendyol,
+  // hepsiburada, n11, getir, yemeksepeti) below as separate, focused entries.
+  {
+    name: 'sahibinden',
+    category: 'general',
+    match: (url) => /^https?:\/\/(www\.)?sahibinden\.com\//.test(url),
+    notes: `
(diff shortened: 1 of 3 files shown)


esokullu:
thank you! this is great! let's add more adapters if we can i'll merge this one. btw we have captcha plugin which can go thru captcha traps, so i don't know if we should take it into account in this PR!? wdyt?

Ohualtex:
@esokullu Thanks, that genuinely means a lot! I'll keep them coming — trendyol, hepsiburada, n11, getir and yemeksepeti are all good candidates, one focused PR each.

On the captcha plugin: I dug into the wiring before answering. When someone has CapSolver enabled, the ACT prompt already grows that `[CAPTCHA SOLVER]` note and the model reaches for `solve_captcha` on its own — so the adapter doesn't need to spell that out for it to kick in.

sahibinden is the awkward case, though. Its wall is DataDome, and `captcha-solver.js` only knows how to detect reCAPTCHA, hCaptcha and Turnstile right now. DataDome is a different animal — a slider/device check, sometimes just a flat IP block — so nudging the agent toward `solve_captcha` here would mostly error out and send it looping, which is exactly what that "stop and tell the user" line is guarding against. That's why I left it as-is, same as the bloomberg adapter does.

If you'd actually like it to walk through DataDome, CapSolver has a dedicated endpoint for it — happy to open a separate issue/PR for that, since it's a real feature rather than an adapter tweak. Just say the word.

esokullu:
Thank you! Yeah, no need for DataDome for now. CONTRIBUTIONS.md calls out Turkish regional adapters as the #1 want, so next up I'd suggest Trendyol, Hepsiburada, n11, Getir, Yemeksepeti — the most-trafficked e-commerce / quick-commerce / food-delivery sites in TR. One PR or a small batch grouped by category both work, whatever's easier. 🙏🙏

PR #243, opened 2026-07-02
fix: accept JSON in file picker and use app locale for speech input

Elijah-hash7:
1. **File picker now accepts JSON** — added `application/json` to the `<input>` accept attribute and the JS type filter in `handleAttachedFiles()`.

2. **Mic uses app locale instead of browser default** — `recognition.lang` now reads from `getLocale()` (the globe icon setting) instead of `navigator.language`, so speech recognition matches the language the user actually selected.

Both Chrome and Firefox builds updated.

### DIFF
--- a/src/chrome/src/agent/agent.js
+++ b/src/chrome/src/agent/agent.js
@@ -9712,6 +9712,8 @@ Rules: no prose intro, no conclusion, no "this screenshot shows...", no layout d
           source: { type: 'base64', media_type: 'application/pdf', data: String(att.dataUrl || '').split(',')[1] || '' },
           ...(att.name ? { title: att.name } : {}),
         });
+      } else if (att.kind === 'text') {
+        blocks.push({ type: 'text', text: `[Attached file: ${att.name || 'file.json'}]\n${att.textContent || ''}` });
       }
     }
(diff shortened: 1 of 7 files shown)


esokullu:
Thanks @Elijah-hash7 ! here are some comments from Github CoPilot, I believe , we should be able to import them as raw txt among other txt like data formats

Elijah-hash7:
Yeah, JSON would just hit the document/PDF path and break. I'll read it as text instead.
will push a fix now thanks!

Elijah-hash7:
Here's what I changed:
- File picker now accepts JSON files added application/json to accept attribute
- JSON files are read as plain text and injected as a text block in the message, instead of going through the PDF/document pipeline so it works with any provider
- All changes mirrored across Chrome and Firefox builds

webbrain-one:
@Elijah-hash7 we seem to have some low and medium level review requests from copilot pls check above https://github.com/webbrain-one/webbrain/pull/243#pullrequestreview-4619007991 🙏

Elijah-hash7:
Addressed the copilot feedback, removed the dead lang fallbacks and added a test for the text attachment path 
Should be good now.

PR #294, opened 2026-07-06
feat: add LocalAI as built-in local provider

Ayush7614:
## Summary
- Add **LocalAI (Local)** as a seventh built-in local provider (`http://localhost:8080/v1`), matching the existing OpenAI-compatible local backend pattern (Jan / vLLM / SGLang).
- Wire LocalAI through Chrome + Firefox: provider defaults, settings card, Load models, onboarding local scan, and docs.
- Extend regression tests for categorization, model listing, and default config presence.
- Fix stale OpenRouter default-config test (`openrouter/free` is the current default on main).

## Test plan
- [x] `npm test` — 661 passed, 0 failed
- [x] Settings → Providers → Local → **LocalAI (Local)** appears with default URL `http://localhost:8080/v1`
- [x] **Load models** lists models when LocalAI is running with OpenAI API enabled
- [x] **Test connection** succeeds against a running LocalAI instance
- [x] Onboarding local scan detects LocalAI when server is reachable

## Notes
LocalAI shares port 8080 with llama.cpp’s default host, but uses the `/v1` OpenAI-compatible path. Users running both should change one server’s port.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Open-source AI browser agent for Chrome and Firefox. Chat with any web page, aut
 - **Continue from Limit** — When the agent hits the step limit, click Continue to keep going
 - **Multi-Provider LLM** — Supports local and cloud models:
   - **WebBrain Cloud 1.0** (cloud, default) — Built-in managed cloud option; no local setup required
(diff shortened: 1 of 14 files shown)


Ayush7614:
cc: @esokullu

PR #309, opened 2026-07-09
providers: add AWS Bedrock (Converse)

Ayush7614:
## Summary
- Add `aws_bedrock` provider implementing the Bedrock Runtime Converse API with SigV4 signing.
- Add Settings fields for region, credentials, and Bedrock model id.
- Mirror implementation and registration across Chrome + Firefox.

## Test plan
- Load unpacked extension.
- Configure AWS Bedrock provider with region + credentials + model id.
- Click **Test Connection**.
- Send a simple prompt in the side panel.


Ayush7614:
cc: @esokullu

PR #320, opened 2026-07-10
fix: auto-detect local provider context window (#106)

### LINKED ISSUE #106: Local model context window can be overstated (16k default vs real 8k window) — expose/auto-detect it
## Summary

The token-aware auto-compaction added in #102 reads the model's window from `provider.contextWindow` (`src/chrome/src/providers/base.js`, mirrored in Firefox). For local backends without an explicit `config.contextWindow`, it returns a conservative **16k** default. But the settings UI does not expose a context-window field, so a user running a local server with a smaller real window (e.g. **8k** — which our onboarding says is workable with Compact mode) gets `contextWindow = 16384`.

Auto-compaction then waits until ~`0.75 × 16384 ≈ 12,288` input tokens before firing — already **past an 8k server's hard limit** — so that setup hits context-overflow / `_emergencyTrim` instead of the smooth token-aware compaction the feature is meant to provide.

Raised by the review bot on #102; deliberately **not** fixed in that PR (it's UI/provider-config work, out of scope for the compaction change).

## Why the default alone can't solve it

It's a genuine trade-off with one fixed number:
- **16k default** (current): good for capable local models (avoids premature compaction), but overstates 8k models → overflow.
- **8k default**: protects 8k models, but over-compacts capable 16k/32k/128k local models, throwing away context unnecessarily.

The real fix is to know each model's **actual** window rather than guess.

## Recommended fix (pick one or both)

1. **Expose a "Context window (tokens)" field** in the per-provider settings UI (next to Base URL / model), persisted into `config.contextWindow`. `base.js` already honors it — this just lets users set the truth. Default the field's placeholder to the category default (16k local / 128k cloud).
2. **Auto-detect the window** from the local server and populate `config.contextWindow`:
   - llama.cpp: `GET /props` → `default_generation_settings.n_ctx` (or top-level `n_ctx`).
   - Ollama: `POST /api/show` → `model_info`'s `*.context_length`.
   - LM Studio: `GET /api/v0/models` exposes `loaded_context_length` / `max_context_length`.
   Fall back to the conservative default when detection fails or isn't supported.

Suggested: do **both** — auto-detect on connect/model-select, and keep the manual field as an override for servers that don't report it.

## Acceptance
- An 8k local model compacts before ~6k input tokens (≈ 0.75 × 8k) instead of waiting for ~12k.
- A capable 32k+ local model is not over-compacted (its real window is used).
- Applies to Chrome and Firefox (both read `provider.contextWindow`).

## Pointers
- `src/chrome/src/providers/base.js` `get contextWindow()` (and Firefox copy) — the default lives here.
- Per-provider settings field definitions: `src/chrome/src/ui/settings.js` (e.g. the existing `baseUrl` field).
- Provider construction / config: `src/chrome/src/providers/manager.js`.

Found during the #102 review sweep.

akmalsyrf:
## Summary

Fixes #106.

Local providers defaulted to a 16k `contextWindow` even when the server was actually running with a smaller window (e.g. 8k). Token-aware auto-compaction then waited until ~75% of 16k (~12k tokens) — past an 8k hard limit — so users hit context overflow / emergency trim instead of smooth compaction.

The Settings "Context window (tokens)" field already existed. This PR adds the missing half of #106: **auto-detect** the real window from the local server on **Test connection** and **Load models**, persist it when appropriate, and fill the Settings field.

### Detection sources

- **llama.cpp** — `GET /props` → `default_generation_settings.n_ctx` / `n_ctx`
- **Ollama** — `GET /api/ps` live `context_length` first (only for the configured model — does not borrow another running model's window), then `POST /api/show` → Modelfile/`parameters` `num_ctx` (string or object). **Does not** use `model_info.*.context_length` or any other top-level `/api/show` field: that's the architecture max, and it overstates the live window — the exact bug this PR fixes. (The issue's original suggestion pointed at `model_info.*.context_length`; this PR intentionally deviates from that for Ollama.)
- **LM Studio** — `GET /api/v0/models` → preferred/loaded model's `loaded_context_length` / `max_context_length` (case-insensitive id match; does not borrow another model's window)

Jan / vLLM / SGLang / LocalAI do **not** auto-detect yet — they keep the manual/default value. Docs call this out explicitly.

### Apply policy

- Refresh when the stored value is missing or still the 16k default
- **Shrink a non-default manual override only from trusted live/runtime detection** (llama.cpp `/props`, Ollama `/api/ps`, LM Studio *loaded* context). Ollama `/api/show` `num_ctx` alone can refresh the 16k default but will not overwrite a deliberate larger override, since it isn't guaranteed to reflect the currently-running allocation.
- Never enlarge a non-default manual override
- Detected/handoff values below the Settings minimum (4k) are **clamped up to 4k** rather than rejected outright — a genuinely small server window (e.g. Ollama's 2k/4k defaults) still replaces the overstated 16k default and stays consistent with the Settings field's own `min: 4096`.

Manual override in Settings still works. Backends that don't report a usable window keep the 16k default. Mirrored in Firefox. Docs updated (EN / zh-CN / fr) — scoped to the three backends that actually auto-detect. Unit tests cover parsers, `num_ctx` string parsing, `/api/ps` model matching, apply policy, and `ProviderManager` wiring (persist / no-borrow / show-only-vs-live-shrink).

## Steps to reproduce (before)

1. Start a local model with an **8k** context, e.g.:

   `llama-server -m your-model.gguf --port 8080 -c 8192`

2. Load the Chrome extension from `src/chrome/`.
3. Settings → llama.cpp → set Base URL `http://localhost:8080`, leave **Context window** at default `16384`.
4. Set llama.cpp as active provider.
5. Start a long Act-mode chat that grows past ~8k input tokens.
6. Observe: compaction does **not** fire near ~6k tokens; the run hits context overflow / emergency trim around the server's 8k limit.

## Steps to verify (after)

1. Same 8k llama.cpp server as above.
2. Settings → llama.cpp → **Test connection** or **Load models**.
3. Confirm **Context window (tokens)** updates to `8192`.
4. Long conversation: auto-compaction should fire before ~`0.65 × 8192` input tokens.
5. Optional: Ollama — with the configured model loaded, Test/Load should prefer `/api/ps` CONTEXT (not the architecture max from `/api/show` `model_info`). If the configured model is **not** the one currently loaded, it must not adopt another running model's window (falls back to `/api/show`).
6. Optional: set Context window manually to `8192`, Save, then Test against a larger server window — the smaller override should stick.
7. Optional: set Context window to `32768` while the server is 8k, then Test:
   - via live detection (llama.cpp `/props` or Ollama `/api/ps`) → shrinks to `8192`
   - via Ollama `/api/show` only (model not in `/api/ps`) → manual `32768` is preserved

## Test plan

- [x] `npm test` — passed (683/683 — parsers, apply policy, and `ProviderManager` wiring for llama.cpp / Ollama / LM Studio)
- [x] Manual: llama.cpp `-c 8192` → Test connection fills `8192`
- [x] Manual: Ollama loaded model → Load models fills live `/api/ps` context (not 128k arch max)
- [x] Manual: Ollama configured model **not** loaded → does not adopt another running model's context
- [x] Manual: LM Studio loaded model → Load models fills `loaded_context_length`
- [x] Manual: long chat on 8k local model compacts before overflow
- [x] Manual: smaller non-default override is not enlarged by Test connection
- [x] Manual: larger non-default override is not shrunk by Ollama `/api/show`-only detection


esokullu:
This is a great and very important feature indeed. Many thanks 🙏 
Looking into it  👀

akmalsyrf:
> This is a great and very important feature indeed. Many thanks 🙏 Looking into it 👀

Thank you, looking forward to contributing more to this repository

PR #348, opened 2026-07-12
feat: add /export-with-traces command

AlsoTheZv3n:
Adds a new **`/export-with-traces`** slash command that downloads the conversation's **tool chain** as Markdown, sourced from the trace store. This is the follow-up you suggested when you reopened this PR: `/export` stays messages-only and is **byte-for-byte untouched** here; this is a separate, additive command.

### Why this replaces the original #348 approach
The first version of this PR tried to make `/export` itself include the tool chain by reading `this.conversations`. That was wrong twice over — it read the wrong source (the live message array is compacted, enriched, and its tool results are `_wrapUntrusted`-wrapped, so `looksFailed`-style parsing was unreliable), and the message never reached the background (raw `chrome.runtime.sendMessage` without `target:'background'` was silently dropped, so it always fell back to the DOM). Both are gone. This version:

- **Sources from the trace store** (`trace/recorder.js` → IndexedDB), an append-only, compaction-immune log whose tool results are the **raw** structured values (pre-truncated by the recorder, never wrapped). That is the correct source for a tool chain.
- **Routes through `sendToBackground('export_traces', …)`** (the `target:'background'` helper), so the message actually reaches the handler.

### What the export contains
Per run, in event order: the user's submitted message as the turn header (`run.userMessage`, pre-enrichment — verified it is the submitted text, not the wrapped/enriched copy), the planner turn (labelled **Planner** and fenced, keyed on the recorder's real `phase:'planner'` — no JSON sniffing; the model emits its plan already wrapped in ```` ```json ````, which the exporter unwraps so it renders as one clean block instead of a broken nested fence), assistant prose, and each tool call as `🔧 name(args) → result`, with `✗` on failed results and the recorder's truncation marker preserved. Example from a live Galaxus run:

~~~
## Turn 1 — find me a lenovo laptop below 1000
_claude-haiku-4-5 · done_

**Planner:**
```json
{"summary":"Search Lenovo <1000","steps":["search","filter price"]}
```
**WebBrain:** Ich suche nach Lenovo Laptops.
- 🔧 `set_field`({"ref_id":"ref_1",…}) → ✗ {"success":false,"error":"ref_id ref_1 not found"}
- 🔧 `extract_data`({"type":"products"}) → ✗ {"error":"Unknown data type"}
- 🔧 `get_accessibility_tree`(…) → …  [recorder-truncated, 20.5kb total]
**WebBrain:** 506 Produkte gefunden.
~~~

### What it does NOT contain (stated in the file's own footer)
Screenshots, notes and vision sub-calls are recorded but **not rendered** — the file ends with a footer saying so and pointing to the Traces page.

### How it relates to the two existing exports
It doesn't overlap either. The **Traces page** exports **JSON**, one run, machine-readable (`schema: webbrain-trace/1`). The **History page** exports **Markdown** of the saved message text (with the trace run-IDs listed, but *not* the tool chain — `recordToMarkdown` in `history.js`). `/export-with-traces` is the piece neither provides: the **tool chain itself, as readable Markdown, for the current conversation.**

### Tracing off
The command depends on **Record traces** being enabled. When it's off (the default) there are no runs to read, so the command shows: *"No traces for this conversation. Enable Record traces in Settings, then run again."* — no silent empty file.

### Restart / reopened conversations (known limit)
`exportTraces` calls `await this._hydrate(tabId)` **before** reading the id, so it repopulates `this.conversationIds` from the `storage.session` mirror. That means an ordinary **service-worker eviction** (the MV3 worker sleeping after ~30s idle) is fully covered — the export still works. The one case it does **not** cover is a **browser restart**, which wipes `storage.session` entirely: `_hydrate` then finds nothing, the id is gone, and the command reports "no traces." I checked whether the sidepanel could recover the old id after a restart — it can't: the History page is read-only, with no path back into the sidepanel. So the read stays the plain (hydrated) `conversationIds.get(tabId)`, and the **browser-restart case is the known limit**, best fixed on the History page (see the open questions).

### Privacy — same surface as the existing trace export
The file carries the same class of data as the Traces-page export (raw page content, typed field values — e.g. `set_field({"text":…})` on a login page could be a password), with the **same redaction posture** (neither redacts), and in fact **less** (no screenshots). Local download only — nothing leaves the browser. **No new data surface:** anyone already comfortable with the trace export is exposing nothing new here.

### Slash-command flags (your "plan slash command flags at large")
I did **not** build a flag system in this PR — it's one command, kept small and reviewable. Happy to take that on separately; my read is a shared parser + a per-command allow-list of flags, but I'd rather agree the shape with you first than smuggle it in here.

### Open questions for you
1. Should `/export`'s output carry a one-line pointer to `/export-with-traces` (discoverability), or keep them fully independent? I did **not** add it — a UX call for you.
2. **The restart gap, as a product question.** A conversation reopened after a restart can't export its trace from the sidepanel — the records are in IndexedDB and the History record even stores the `conversationId`, but the sidepanel can't reach it. The History page already resolves a record's trace runs (it renders a **Linked Traces** panel via `traceRunsForRecord()`), so the natural fix is an **"Export tool chain" button on the History page**, reusing this exact serializer — or a "resume in chat" action that rehydrates the id. Happy to take either as a separate follow-up if it's useful.

### Tests / parity / i18n
- `node test/run.js`: this PR's serializer suite passes with **0 regressions**. (The tree reports 774/1 after rebasing onto current `main` — the one failure, `repository changelog retains complete, non-empty release history`, is **pre-existing on `upstream/main` itself**: package.json is at 23.0.4 while the newest CHANGELOG entry is 23.0.2. This PR touches neither `CHANGELOG.md` nor `package.json`, so it's left for the release process.) The new suite covers event ordering, `✗` on failed results, recorder-truncation, screenshot omission, planner label + `` ```json `` fence-unwrap, and the honest footer, plus a chrome/firefox byte-parity assertion on the serializer.
- Chrome + Firefox kept identical (`trace-export.js` is byte-identical across builds; the locale parity test passes).
- **Discovery:** the autocomplete description (`sp.slash.export_traces`) is translated in **all 16 locales**. The `/help` listing (`sp.help_html`) is updated in **English only, deliberately** — the non-English `help_html` strings are already abbreviated (15 `<code>` entries vs English's 38) and not kept in sync, so English-only matches the existing precedent rather than half-syncing one stale string.
- **The 15 non-English translations are machine-generated** (no `de.js` to cross-check against) — please have a native speaker spot-check before relying on them; they contain no interpolation placeholders, so they're parity-safe.
- Additive by construction: **315 insertions, 2 deletions** — the two deletions are the English `sp.help_html` line (both builds), replaced to list the new command. The `/export` handler is **byte-identical** to upstream.

New file: `src/{chrome,firefox}/src/agent/trace-export.js` (pure serializer, unit-tested without a DOM/IndexedDB).


esokullu:
@AlsoTheZv3n thank you. Please confirm these are all addressed.

esokullu:
btw have you seen the Traces feature in the Settings panel and the Traces screen? It has Export feature too. I wonder if it is the behaviour you're looking for.

AlsoTheZv3n:
Thanks — I went through all ten. They're valid, and two of them decide it, so I'm closing this rather than patching around them.

**The routing comment is fatal on its own.** I sent `chrome.runtime.sendMessage({ action, tabId })` without `target: 'background'`, so the central listener drops it and `/export` always falls through to the DOM path. That's worse than a no-op: my fallback banner ("restored from a previous session") would then print on *every* export, which is false. My tests were green because they exercised the pure serializer in isolation — they never touched the routing, the `<untrusted_page_content>` wrapping, or the enriched user turns — and I never ran `/export` in the loaded extension after wiring it. My mistake, and a clear one.

**The compaction comment kills the approach, not just the code.** `this.conversations` is the model's working context, not a run transcript — `_manageContext` summarises and drops older messages, so the export would be least complete on exactly the long, derailing runs it was meant to document. "Every call, in order" can't come from that source, and getting it there would mean maintaining a second, lossy transcript.

**On Traces: you're right.** That's the purpose-built, structured, compaction-immune log, and it's what I should have used. I did find `trace/recorder.js` while investigating and set it aside because tracing is off by default — I should have read "off by default" as a discoverability question, not as a reason to reconstruct the chain from the wrong source.

What's left of the finding is small but real, and it's how I ended up here: `/export` produces a file that *looks* complete while silently omitting every tool call. I exported a 38-step run that had hit an Akamai wall and retry-looped, and the markdown read like a clean two-message exchange. Two coherent ways to address that — your call, and I'd rather ask than open another PR:

- **(a)** `/export` stays messages-only, but says so and points at Traces (a few lines, no new surface). I'd pick this.
- **(b)** `/export` carries the tool chain, sourced from the trace store when tracing is on — not from `conversations`.

Happy to open a small PR for either, and equally happy to leave `/export` as it is if you consider it working as intended. Closing this one regardless. Thanks for the review.

esokullu:
I think /export should stay messages-only but we may have a second /export-with-traces command for, as the name suggests, export with traces 
btw of course it would be a better design if we had slash command flags too, like:
/export --with-traces instead of /export-with-traces but it's not a pattern we've used yet, but must have used 

so feel free to turn this into /export-with-traces
or even take on a larger task (perhaps as a second step) and plan slash command flags at large (we may need it elsewhere too)

Thanks regardless!

src/chrome/src/ui/sidepanel.js
```diff
@@ -388,6 +389,7 @@ const OUT_OF_BAND_SLASH_COMMANDS = new Set([
   '/list-schedules',
   '/dangerously-skip-permissions',
   '/screenshot',
+  '/export-with-traces',
```
esokullu, line 392: **[suggestion]** `/export-with-traces` is correctly added to `OUT_OF_BAND_SLASH_COMMANDS` (and the same change is mirrored in Firefox), so it can run while an agent run is in progress. However, `sp.slash.busy_only_oob` in every locale (including `en.js`) still lists only `/export` and not `/export-with-traces`. Users who hit the busy toast will not know the new command is allowed, and the static locale string checked by `test/run.js` will keep this drift locked in until both the strings and the test are updated together.

**Suggestion:** Append `/export-with-traces` to `sp.slash.busy_only_oob` in Chrome and Firefox locales (at least `en`, ideally all languages that already list `/export`), and update the exact-string assertion(s) in `test/run.js` that pin that copy.

src/chrome/src/agent/agent.js
```diff
@@ -6307,6 +6308,38 @@ Rules: no prose intro, no conclusion, no "this screenshot shows...", no layout d
...
+   * 'no-traces' (tracing off / nothing recorded) so the UI can say so instead of
+   * downloading an empty-but-official-looking file.
+   */
+  async exportTraces(tabId) {
+    await this._hydrate(tabId);
+    const conversationId = this.conversationIds.get(tabId);
+    if (!conversationId) return { ok: true, markdown: null, turnCount: 0, reason: 'no-conversation' };
+    let runs;
+    try {
+      const all = await trace.listRuns();
```
esokullu: **[suggestion]** `exportTraces` calls `trace.listRuns()` with no options, so it inherits the recorder default `limit: 500` (newest-first across *all* conversations). Filtering by `conversationId` happens only after that global cap. If the user has many traced runs on other tabs/chats, older turns of the *current* conversation can be silently omitted from the export while still producing a successful-looking Markdown file. The Traces UI has the same limit, but a command named for “the tool chain” of this conversation is more likely to be trusted as complete.

**Suggestion:** Prefer a conversation-scoped query if one can be added to the recorder, or call `listRuns({ limit: … })` with a higher cap and/or page until no more matching runs are found. At minimum, surface in the success message or footer when the export may be truncated (e.g. when `all.length === limit`).

src/chrome/src/agent/trace-export.js
```diff
@@ -0,0 +1,114 @@
...
+function renderResult(result) {
+  if (result == null) return { text: '(no result recorded)', failed: false };
+  if (typeof result === 'object' && result._truncated) {
+    return {
+      text: `${truncate(oneLine(String(result.head ?? '')), RESULT_LIMIT)}  [recorder-truncated, ${humanSize(result.length || 0)} total]`,
+      failed: false,
+    };
+  }
+  const failed = typeof result === 'object' ? (result.success === false || !!result.error) : false;
+  const s = typeof result === 'string' ? result : JSON.stringify(result);
```
esokullu: **[suggestion]** `renderResult` / `stringifyArgs` use bare `JSON.stringify`. The recorder only truncates large results when stringify succeeds; on failure it stores the original value, and IndexedDB can persist non-JSON-serializable / circular structures via structured clone. If such a tool result is later exported, `tracesToMarkdown` throws, and `exportTraces` does not wrap that call in try/catch—so the side panel falls into the generic error path with little context.

**Suggestion:** Guard `JSON.stringify` (and optionally the whole `tracesToMarkdown` call in `exportTraces`) with try/catch and fall back to `String(result)` / `'(unserializable result)'`, returning `{ ok: false, error }` only for unexpected hard failures.

src/chrome/src/agent/agent.js
```diff
@@ -6307,6 +6308,38 @@ Rules: no prose intro, no conclusion, no "this screenshot shows...", no layout d
...
+        .filter((r) => r && r.conversationId === conversationId)
+        .sort((a, b) => (a.startedAt || 0) - (b.startedAt || 0));
+    } catch (e) {
+      return { ok: false, error: String((e && e.message) || e) };
+    }
+    if (runs.length === 0) return { ok: true, markdown: null, turnCount: 0, reason: 'no-traces' };
+    const withEvents = [];
+    for (const run of runs) {
+      let events = [];
+      try { events = await trace.getRunEvents(run.runId); } catch (e) { events = []; }
```
esokullu: **[suggestion]** `getRunEvents` failures are swallowed per run (`events = []`) with no signal to the UI. A partial IDB/read failure can yield a download that still has turn headers (and `turnCount > 0`) but is missing the tool chain the user asked for, with success toast “Conversation exported.”

**Suggestion:** Track failed event loads; if any (or all) fail, return `ok: false` or attach a warning into the Markdown/header and a distinct UI message so the user does not trust a hollow export.

src/chrome/src/ui/sidepanel.js
```diff
@@ -3656,6 +3658,40 @@ async function parseSlashCommands(text, tabId = currentTabId) {
...
+    a.href = url;
+    a.download = `webbrain-traces-${Date.now()}.md`;
+    document.body.appendChild(a);
+    try {
+      a.click();
+    } finally {
+      a.remove();
+      setTimeout(() => URL.revokeObjectURL(url), 7000);
+    }
+    addPersistentSlashMessage(t('sp.export.done'));
```
esokullu: **[nit]** On success the UI reuses `t('sp.export.done')` (“Conversation exported.”). This export is deliberately *not* the chat transcript—it is the tool-chain/traces view—so the confirmation is slightly misleading and may confuse users comparing `/export` vs `/export-with-traces`.

**Suggestion:** Add something like `sp.export_traces.done` (“Tool chain exported.” / “Traces exported.”) in Chrome and Firefox locales and use it here (and in the Firefox side panel mirror).

src/chrome/src/ui/sidepanel.js
```diff
@@ -3656,6 +3658,40 @@ async function parseSlashCommands(text, tabId = currentTabId) {
...
+      res = await sendToBackground('export_traces', { tabId });
+    } catch (e) {
+      addPersistentSlashMessage(`${t('sp.export_traces.error')} (${e?.message || e})`);
+      return '';
+    }
+    if (!res?.ok) {
+      addPersistentSlashMessage(`${t('sp.export_traces.error')} (${res?.error || 'unknown error'})`);
+      return '';
+    }
+    if (!res.markdown || res.turnCount === 0) {
```
esokullu, line 3675: **[nit]** The empty path treats every zero-turn outcome as `sp.export_traces.none` (“No traces for this conversation. Enable Record traces…”). `exportTraces` already distinguishes `reason: 'no-conversation'` vs `'no-traces'`, but the UI ignores `reason`. “Enable Record traces” is the wrong guidance when there is simply no conversation id for the tab yet.

**Suggestion:** Branch on `res.reason` (or separate `ok` payloads) so no-conversation can say there is nothing to export yet, while no-traces keeps the tracing-settings hint.

test/run.js
```diff
@@ -1633,6 +1641,65 @@ test('GitHub Enterprise does not match github adapter (strict)', () => {
   assert.equal(a, null);
 });
 
+// ────────────────────────────────────────────────────────────────────────
+// /export-with-traces serializer tests
+// ────────────────────────────────────────────────────────────────────────
+
+console.log('\ntrace export');
+
```
esokullu, line 1649: **[suggestion]** New tests cover the pure serializer well (ordering, planner fencing, failure markers, truncation, screenshot omission, Chrome/Firefox parity). There are no static wiring checks of the kind this suite already uses for slash commands—e.g. that both side panels register `/export-with-traces` in `SLASH_COMMANDS` and `OUT_OF_BAND_SLASH_COMMANDS`, parse it *before* `/export`, call `export_traces`, and that both backgrounds handle `case 'export_traces'`. A future refactor could drop the handler or reintroduce the `/export\b` prefix bug without failing CI.

**Suggestion:** Add small source-level assertions (similar to the existing OOB and `/export` blob-URL tests) for command registration, handler order, and the background message name on Chrome and Firefox.

esokullu (commented):
## Summary

PR #348 adds `/export-with-traces`, a slash command that serializes the current conversation’s tool chain from the IndexedDB trace store into a Markdown download. The architecture is sound: a pure `tracesToMarkdown` helper (Chrome/Firefox mirrored), agent-side hydration + conversationId filtering, background message wiring, UI ordering before `/export` (because `/export\b` would otherwise match), and unit tests for the serializer. Dominant residual risks are incomplete busy-state UX copy, silent truncation via `listRuns`’s 500-run default, and thin wiring/test coverage outside the pure serializer.

## Issue counts by severity

- bugs: 0
- suggestions: 5
- nits: 2

esokullu:
Pushed review follow-ups on `2f28680c`:

1. **busy_only_oob** — `/export-with-traces` listed in all Chrome/Firefox locales + tests
2. **listRuns scope** — conversation-scoped listing so other chats don’t starve this conversation’s export; truncation note when the per-conversation cap is hit
3. **JSON.stringify** — safe fallback in `trace-export.js` (Chrome/Firefox)
4. **getRunEvents failures** — reject on IDB error; export returns `partial` / hard-fails if all loads fail; notes in markdown + distinct UI toast
5. **Success toast** — `sp.export_traces.done` (not conversation export copy)
6. **Empty reasons** — UI branches on `no-conversation` vs `no-traces`
7. **Wiring tests** — static checks for slash order, OOB set, `export_traces` background handler

Please re-test `/export-with-traces` while a run is busy and with tracing off / empty tab.

esokullu:
Thanks @AlsoTheZv3n it's merged!

Issue #357, opened 2026-07-13
new skills


esokullu:
OTP / verification-code helper: read recent emails/SMS-like page content and extract likely one-time codes safely.

Temporary file-share skill: use short-lived, no-account file transfer services for non-sensitive files, with expiry warnings.

Public-page archiver skill: save or snapshot non-sensitive public pages the user may need later.

Coupon / trial cleanup skill: track free-trial signup details and remind the user to cancel before renewal.

Alias email guidance skill: for important-ish accounts, prefer user-owned aliases like iCloud Hide My Email, Fastmail masked email, Proton aliases, or SimpleLogin instead of disposable inboxes.

Account-risk classifier skill: before signup, classify whether disposable email is appropriate or whether the user should use a durable address.

alphatrann:
@esokullu can I contribute to this?

esokullu:
sure that would be great
perhaps start with this one: Temporary file-share skill: use short-lived, no-account file transfer services for non-sensitive files, with expiry warnings.

make sure there is a service which requires no api, no auth, publicly accessible
we need something like https://github.com/webbrain-one/webbrain/blob/3a081e224194ca0ee601a7e5bcce523b040629d3/src/chrome/skills/disposable-email-mailtm.md
and its ff equivalent 
https://github.com/webbrain-one/webbrain/blob/3a081e224194ca0ee601a7e5bcce523b040629d3/src/firefox/skills/disposable-email-mailtm.md

alphatrann:
@esokullu sounds good, I'll scope this to just the temporary file-share skill for now. Planning to model it on `disposable-email-mailtm.md`, mirrored for chrome + firefox. I'm looking into [Litterbox](https://litterbox.catbox.moe/) since it satisfies what you expected (no api, no auth, publicly accessible)

Will open a draft PR once I have something to look at.

alphatrann:
@esokullu I've implemented and verified e2e the temporary file-share skill with litterbox. Everything worked like a charm.

`npm test` has one failing test: `repository changelog retains complete, non-empty release history` asserts the changelog's top section matches `package.json`'s version. I confirmed via a clean worktree of upstream `main` that this same test already fails there (806 passed/1 failed, identical), so it's pre-existing drift on the base branch, not something this PR caused or should silently fix.

I'll continue to add the remaining as you requested :)

esokullu:
fixed @alphatrann  and merged. thanks!

PR #367, opened 2026-07-14
feat(skills): add temporary file-share skill (Litterbox)

alphatrann:
## Summary
- Adds an opt-in packaged skill (Chrome + Firefox) that shares a single non-sensitive file through Litterbox, a short-lived no-account host, with user-selectable retention (1h/12h/24h/72h) and public-link/expiry warnings.
- Registered in `PACKAGED_SKILL_SOURCES` but not `DEFAULT_SKILL_SOURCES`, so it's opt-in like the existing Mail.tm skill.
- Drives the Litterbox upload form via `upload_file` rather than its API, so no `/allow-api` override is needed and file bytes never pass through the configured LLM provider.
- Partial progress on #357 (temporary file-share item only; the other proposed skills in that issue — OTP helper, page archiver, coupon cleanup, alias email guidance, account-risk classifier — are not part of this change).

## Test plan
- [x] `npm test` (unit + injection-corpus) — 780 unit tests + 60/60 injection-corpus checks passing, run 3x clean
- [x] Verified Litterbox's live retention options (1h/12h/24h/72h ids, displayed as "1 Hour/12 Hours/1 Day/3 Days") and 1GB size cap match the skill's claims
- [x] Verified `upload_file`/`clarify` tool contracts and Mid/Full-tier + Act-mode gating against `tools.js`/`permission-gate.js`
- [x] Verified the flow cannot trigger `/allow-api` (gate only fires for `fetch_url`/`research_url` mutations, which this skill never calls)
- [x] Manually loaded the unpacked extension, enabled the skill in Settings → Skills, and completed a real Act-mode share end-to-end — link returned with absolute expiry, no `/allow-api` prompt

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This changelog was generated from the repository Git history and release tags. V
 ### Added
 - Added universal `<command> --help` support for slash commands, returning the selected command's usage, description, and available options directly in chat.
 - Published the WebMCP integration blog post.
+- Added an opt-in packaged Litterbox temporary file-share skill for Chrome and Firefox, available from Settings with explicit `clarify` confirmation, public-link and absolute-expiry warnings, non-sensitive-file-only rules, and visible provider attribution. Uploads go through the Litterbox page with `upload_file`, so no `/allow-api` override is needed and file contents never reach the LLM provider.
 
 ### Fixed
 - Made Enter accept highlighted slash-command flag completions such as `/schedule --list` without prematurely executing the parent `/schedule` command.
@@ -16,6 +17,7 @@ This changelog was generated from the repository Git history and release tags. V
 
 ### Tests
 - Added Chrome and Firefox regression coverage for command-specific help, flag autocomplete ordering and completion, invalid mixed help flags, and trusted toast rendering.
+- Added packaged-skill catalog and Litterbox safety/expiry coverage.
 
 ## [23.1.0] - 2026-07-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,19 @@ All notable changes to WebBrain are documented in this file.
 
 This changelog was generated from the repository Git history and release tags. Versions without a Git tag are inferred from version-bump commits and the current `package.json` / browser manifest versions.
 
+## [Unreleased]
+
+### Added
+- Added an opt-in packaged Litterbox temporary file-share skill for Chrome and Firefox, available from Settings with explicit `clarify` confirmation, public-link and absolute-expiry warnings, blocked-file preflight checks, browser-specific upload limits, and visible provider attribution. Uploads go through the Litterbox page with `upload_file`, so no `/allow-api` override is needed and file bytes never reach the LLM provider.
+
+### Tests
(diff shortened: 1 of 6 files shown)


esokullu:
Additional review findings to address:

- **[P2] Cap Firefox uploads at 25 MB in the skill** — `src/firefox/skills/temporary-file-share-litterbox.md:7`
  Firefox rejects picker and `downloadId` uploads over `UPLOAD_MAX_BYTES` (25 MB), so the skill must document and preflight the browser-specific limit instead of advertising 1 GB.

- **[P2] Wait for Dropzone instead of submitting again** — `src/chrome/skills/temporary-file-share-litterbox.md:40`
  Litterbox uses Dropzone with `autoProcessQueue: true`; `upload_file` starts the transfer via the file input change event. The workflow should wait for the result (for example, `.responseText`) rather than search for a separate submit action.

- **[P2] Disclose the Chrome absolute path to the provider** — `src/chrome/skills/temporary-file-share-litterbox.md:24`
  With the Chrome `filePath` input, the absolute local path is included in the model/tool transcript. The bytes bypass the provider, but path metadata can expose usernames or project directories. Prefer `downloadId` where possible and disclose this metadata.

- **[P2] Reject file types Litterbox does not accept** — `src/chrome/skills/temporary-file-share-litterbox.md:7`
  Preflight should reject the unsupported `.exe`, `.scr`, `.cpl`, `.doc*`, and `.jar` files instead of entering a workflow that cannot succeed.

- **[P2] Record the feature under its actual release** — `CHANGELOG.md:12`
  The feature was added after v23.1.5 but currently appears under the historical 23.1.2 section. Move it to the current or unreleased release section.

esokullu:
Addressed the review findings in **4cdc12fa** (Chrome + Firefox skill docs + tests):

1. **[bug] Upload wait reliability** — Step 8 now uses `wait_for_element` with multi-minute `timeout`s (tens of minutes for large Chrome uploads), forbids relying on `wait_for_stable` alone (20s cap), and requires re-reading until `.responseText` contains a non-empty `https://litter.catbox.moe/` URL (or a visible `.dz-error-message`).

2. **[bug] Firefox preflight vs picker** — Split guidance: when `downloadId`/`list_downloads` metadata exists, apply blocked-type + 25 MB checks before confirmation; on the picker path, confirm public-link/IP/expiry risks first, then report post-attach size/type failures without inventing a link.

3. **[suggestion] Expiry UI labels** — Explicit mapping: 1h → **1 Hour**, 12h → **12 Hours**, 24h → **1 Day**, 72h → **3 Days**; absolute-expiry math still uses the duration.

4. **[suggestion] URL allowlist** — Accept only `https://` URLs on host `litter.catbox.moe`, taken only from `.responseText`.

5. **[suggestion] IP logging** — Pre-upload `clarify` now discloses that Litterbox stores the uploader's IP (per their FAQ).

6. **[nit] Browser-specific upload notes** — Chrome skill keeps `downloadId`/`filePath` + path-metadata disclosure; Firefox skill keeps `downloadId`/picker + 25 MB (no cross-browser bullets).

Regression asserts added in `test/run.js` for host allowlist, IP disclosure, label mapping, wait guidance, and browser-specific paths.

PR #415, opened 2026-07-19
agent: add adapters for Bilibili and Xiaohongshu

alectimison-maker:
Adds mirrored Chrome and Firefox adapters for Bilibili and Xiaohongshu.

Prevents a real interaction failure where custom `div`/`span` controls for Like, Follow, Favorite, Comment, and Publish were absent from the accessibility tree and interactive-element list. Bilibili's open-shadow-root comment controls are now discoverable as stable accessibility refs, and Xiaohongshu's wrapper controls receive semantic action names.

Validation:
- `node test/run.js`: 999 passed; 2 unrelated pre-existing baseline failures remain (CHANGELOG version and localized Tweet recommendation label)
- `node test/security/injection-corpus.mjs`: 60/60 passed
- Live Bilibili DOM/accessibility verification for Like, Coin, Favorite, Share, Follow, Login, and Reply
- Synthetic Xiaohongshu DOM verification for semantic discovery and text-based clicking

### DIFF
--- a/src/chrome/src/agent/adapters.js
+++ b/src/chrome/src/agent/adapters.js
@@ -131,6 +131,40 @@ function isFacebookInfiniteScrollUrl(url) {
   return /^\/(?:watch|groups|marketplace|search)(?:\/|$)/.test(path);
 }
 
+function isBilibiliUrl(url) {
+  const parts = adapterUrlParts(url);
+  if (!parts) return false;
+  const hostname = normalizedHostname(parts.parsed.hostname);
+  return hostname === 'bilibili.com' || hostname.endsWith('.bilibili.com');
+}
+
+function isBilibiliInfiniteScrollUrl(url) {
+  const parts = adapterUrlParts(url);
+  if (!parts || !isBilibiliUrl(url)) return false;
+  const hostname = normalizedHostname(parts.parsed.hostname);
+  if (hostname === 't.bilibili.com') return true;
+  if (hostname === 'bilibili.com' && parts.path === '/') return true;
+  return hostname === 'space.bilibili.com'
+    && /^\/\d+\/dynamic(?:\/|$)/.test(parts.path);
+}
+
+function isXiaohongshuUrl(url) {
+  const parts = adapterUrlParts(url);
+  if (!parts) return false;
+  const hostname = normalizedHostname(parts.parsed.hostname);
+  return hostname === 'xiaohongshu.com' || hostname.endsWith('.xiaohongshu.com');
+}
+
+function isXiaohongshuInfiniteScrollUrl(url) {
+  const parts = adapterUrlParts(url);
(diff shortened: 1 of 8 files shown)


esokullu:
Short answer: No — _siteInteractionText is not a better alternative; it already contains that exact fallback as its last line. Replacing the inline expression with a call to _siteInteractionText(el) would be a strict improvement in this context, because the helper does everything the inline expression does and more.

What I found in src/chrome/src/content/content.js (PR branch feat/bilibili-xiaohongshu-adapters):


js
Copy
function _siteInteractionText(el) {
  try {
    const descriptor = window.__wbSiteInteractions?.describe?.(el);
    if (descriptor?.name) return descriptor.name;
  } catch {}
  return (el?.innerText || el?.value || el?.placeholder || el?.title || el?.ariaLabel || '').trim();
}



So the helper is literally:

1. First, ask the site-adapter layer (window.__wbSiteInteractions?.describe?.(el)) for a site-specific name. This is the new path the PR is introducing — adapters (Bilibili, Xiaohongshu, and the existing ones) can return a curated descriptor.name that reflects what the element actually means on that site (e.g. a "一键三连" button, a "弹幕" toggle, a "选集" episode pill). The inline expression has no way to reach this.
2. Fallback, return the same (innerText || value || placeholder || title || ariaLabel).trim() chain — byte-for-byte identical to what you wrote.

Why the helper is better here, not just equivalent:

- Adapter-aware naming. On Bilibili/Xiaohongshu the visible text is often Chinese, truncated, or wrapped in nested spans that innerText collapses oddly. The adapter's describe() can return a stable, model-friendly label (e.g. English gloss + original) that the inline expression cannot produce.
- Single source of truth. getInteractiveElements, _staleIndexError, and the click-by-text auto-scroll retry all already call _siteInteractionText(el). If a third site adds an adapter, every consumer picks it up automatically; the inline expression would silently bypass it.
- Safer property access. The helper uses el?.innerText || el?.value || … (optional chaining + null-safe), whereas the inline expression you wrote uses el.innerText || el.value || … — if el is ever null/undefined here, the helper returns '' and the inline expression throws.
- Identical fallback cost. When no adapter is registered (the common case on non-adapted sites), the helper returns exactly the same string the inline expression would, with one extra function-call frame — negligible.

One caveat worth flagging in the review: the helper's fallback order is innerText → value → placeholder → title → ariaLabel. That matches your inline expression, but note that el.ariaLabel is the IDL property (reflects aria-label and aria-labelledby resolution in some browsers, and is empty for elements without those). If you specifically wanted the raw getAttribute('aria-label') string, the helper's last term differs slightly. In practice for click-target text this is fine and usually preferable, but it's the one place the two are not byte-identical.

Recommendation: replace the inline expression with _siteInteractionText(el) in this context. It's strictly more capable, already used by the surrounding code, and the fallback is identical.

Issue #305, opened 2026-07-08
WebMCP integration


domfarolino:
Hi! While working on Gemini-in-Chrome (which does not use a local model, but has the same kind of actuation capabilities as WebBrain), we and many partners found that actuation of web content can often be done _much_ faster and much more accurately if the developer gets a chance to expose core site functionality to an agent in the form of "tools" it can understand.

Through this, we [developed WebMCP](https://github.com/webmachinelearning/webmcp) which you may have heard of. Has there been any discussion of adding WebMCP support to WebBrain? This might be as easy as simply finding WebMCP tools through CDP and loading them into the model's context, to make it easier to actuate the site. I'd love to discuss a possible integration more, let me know what you think!

esokullu:
This is really interesting! WebMCP could be a game-changer for browser agents like WebBrain.

The idea of exposing web app functionality as structured tools that agents can discover and invoke directly is compelling. Instead of relying solely on DOM manipulation and accessibility trees, we could have a more semantic layer where developers explicitly define what actions are available and how to invoke them.

A few thoughts:

**Benefits:**
- More reliable than DOM scraping - tools are explicitly defined with schemas
- Better performance - direct function calls vs. multiple DOM interactions
- Clearer intent - developers specify what operations make sense for agents
- Could reduce brittleness when UIs change

**Questions to explore:**
- How would tool discovery work? Would there be a standard manifest or registry?
- Security model - how do we prevent malicious sites from exposing harmful tools?
- Granularity - should tools be high-level (e.g., "create_issue") or low-level (e.g., "click_button")?
- How does this interact with existing approaches like accessibility trees?

I'd love to see a proof-of-concept integration. Even a simple example showing how WebBrain could discover and use WebMCP tools from a test site would be valuable.

@domfarolino have you thought about how this would work with the permission model? Users would need to understand what tools a site is exposing and consent to their use.

domfarolino:
These are great questions, and we've been exploring all of them while developing WebMCP.

> How would tool discovery work? Would there be a standard manifest or registry?

As of now, it's up to the agent to listen for `toolchange` events, which are fired when tools are registered and unregistered. We're considering integration [with ARD](https://agenticresourcediscovery.org/), which would give a site the ability to declarative express the list of tool it supports (granularity and URL patterns/scope have not been discussed yet; we're working through the initial discussions). Declarative discovery is something we're very interested in though.

> Security model - how do we prevent malicious sites from exposing harmful tools?

[...]

PR #416, opened 2026-07-20
feat: add experimental WebMCP fast path via Chrome CDP

alectimison-maker:
## Summary

This pull request adds an experimental, opt-in WebMCP fast path to WebBrain. When the user enables **Experimental WebMCP** under Settings → General → Advanced, supporting Chrome pages let the agent discover and invoke structured capabilities registered by the page through Chrome DevTools Protocol instead of relying exclusively on screenshots, accessibility-tree targeting, or DOM interaction. The setting is off by default, so ordinary runs do not receive WebMCP tool schemas or prompt guidance.

The existing browser interaction stack remains the fallback for pages that do not expose WebMCP tools.

## Motivation

WebBrain already supports visual, accessibility-tree, DOM, and limited observed-API interaction paths. Those approaches remain important, but they can be inefficient or brittle when a site offers an explicit machine-readable operation. WebMCP provides a browser-native way for a cooperating page to publish structured tool names, descriptions, input schemas, and callbacks while preserving the user's active browser session and page state.

This implementation follows the experimental Chrome CDP `WebMCP` domain and deliberately treats the page as an untrusted security principal.

### Chrome CDP integration

- Added per-tab WebMCP sessions backed by `WebMCP.enable` and `WebMCP.disable`.
- Maintains a live catalog from `WebMCP.toolsAdded` and `WebMCP.toolsRemoved` events.
- Invokes registered callbacks through `WebMCP.invokeTool` and consumes asynchronous `WebMCP.toolResponded` events.
- Handles the protocol race where a response event can arrive around the command response.
- Cancels pending invocations on timeout, user stop, session cleanup, or debugger detach.
- Reports dispatched failures as outcome-unknown because the page may have changed state before returning an error.

### Agent tools and routing

- Added a Chrome-only **Experimental WebMCP** toggle under Settings → General → Advanced; it defaults off.
- While disabled, WebMCP tool schemas and prompt guidance are omitted from model requests. Turning it off also closes active WebMCP CDP sessions.
- Added `list_webmcp_tools` for bounded, paginated capability discovery.
- Added `execute_webmcp_tool` for invoking a catalog entry through an opaque `wmcp_*` identifier.
- When enabled, exposes discovery in Ask, Act, and Dev on supported Chrome tool surfaces.
- Restricts execution to Act/Dev and Mid/Full tiers; Compact remains intentionally narrow.
- Marks execution as state-changing and navigation-prone so normal screenshot, navigation, loop, and completion-verification behavior still applies.
- Preserves existing accessibility-tree, DOM, and screenshot tools as the fallback path.

### Security and trust boundaries

- Treats WebMCP names, descriptions, schemas, annotations, frame URLs, outputs, exceptions, and errors as page-supplied untrusted content.
- Routes discovery and execution results through the existing untrusted-content wrapper to prevent prompt-boundary breakout.
- Uses opaque local IDs instead of page-controlled tool names as model-call handles.
- Does not trust page-authored `readOnly` annotations as an authorization boundary.
- Allows Ask mode to list tools but never invoke them.
- Requires Act/Dev, fresh per-call confirmation, and the normal capability permission for every invocation.
- Charges permission to the actual registration-frame host rather than the top-level page host.
- Fails closed for missing, opaque, sandboxed, or non-HTTP(S) frame origins.
- Revalidates the frame ID and effective security origin immediately before dispatch, preventing a navigated iframe from borrowing a previously approved host grant.
- Bounds the catalog to 200 tools, returns at most 25 tools per page, sanitizes page-controlled schemas and outputs, and rejects oversized or non-serializable inputs before dispatch.

### Browser compatibility

- Chrome receives the experimental CDP implementation behind the default-off opt-in.
- Firefox keeps mirrored schemas/security classifications for parity but does not expose the tools at runtime and returns an explicit unsupported/no-dispatch result if called internally.

### Documentation

Updated the README and architecture, privacy/data-flow, prompt-injection, security-model, and testing documentation to describe:

- where to enable WebMCP and its default-off behavior;
- when WebMCP is available;
- what data enters model context;
- the untrusted-page boundary;
- permission and confirmation behavior;
- Chrome/Firefox differences;
- how to run the real-browser smoke test.

### Main unit suite

`node test/run.js`

- 1,104 passed
- 0 failed
- 1,104 total
- All WebMCP regression tests pass, including the default-off model-context gate and mandatory per-invocation confirmation/frame-host permission under bypass modes.

The WebMCP coverage includes:

- feature, mode, and provider-tier gating;
- opaque IDs, pagination, and frame mapping;
- asynchronous invocation and early response races;
- untrusted-content boundary neutralization;
- hostile catalog and input limits;
- unsupported-domain cleanup;
- timeout, cancellation, and outcome-unknown behavior;
- permission metadata ownership;
- Ask-mode denial even for page-declared read-only tools;
- cross-origin frame permission scoping;
- opaque-origin and navigation-race fail-closed behavior;
- Firefox unsupported behavior.

### Prompt-injection corpus

`node test/security/injection-corpus.mjs`

- 60/60 checks passed across Chrome and Firefox.

### Existing browser fixtures

`npm run test:fixtures`

- 114 passed
- 0 failed

### Real Chrome WebMCP smoke test

`python test/webmcp-e2e.py` against the included local fixture, using Chrome 150 with `WebMCPTesting,DevToolsWebMCPSupport` enabled:

- discovered two page-registered tools through CDP;
- transported the declared input schema;
- completed an asynchronous structured invocation;
- verified the resulting page UI state;
- propagated a predictable JavaScript exception;
- observed dynamic tool unregistration.

### Packaging and static validation

- `npm run build:zip` successfully produced Chrome, Edge, and Firefox packages.
- JavaScript syntax checks passed for all touched agent, CDP, permission, and tool files.
- Python compilation passed for the WebMCP smoke test.
- `git diff --check` passed.

### CI harness

`npm run test:ci`

- 9 scenarios validated
- cloud capture test passed

## Local validation

No local test failures remain after merging current `main`.

## Notes

WebMCP is experimental and only accelerates cooperating pages that register tools. It is not a general mechanism for reading arbitrary backend databases or bypassing site authorization. When the API or a relevant tool is unavailable, WebBrain continues through its established accessibility-tree, DOM, screenshot, and observed-API paths.


webbrain-one:
wow thank you this is amazing! we've been thinking about this: https://github.com/webbrain-one/webbrain/issues/305

esokullu:
Because every token counts, especially with smaller local models, I'd like to land this as an explicitly opt-in setting under Settings > Advanced rather than enabling it by default. That keeps the default surface lean for users on tight token budgets, while still making the fast path available for those who want it.

I do hope WebMCP gains ground soon — we'd love to be an early cheerleader. If there's an official icon or support announcement we can use, happy to add it to the website. I didn't see one linked in the PR, so feel free to point me to it.

PR #440, opened 2026-07-21
fix(webmcp): discover pre-registered iframe tools

alectimison-maker:
## Summary

- recover WebMCP tools registered before `WebMCP.enable` from same-process child frames
- recursively aggregate out-of-process iframe targets and route invocation, cancellation, response, and removal events through the owning CDP session
- derive permission URLs from each target's trusted frame tree and clean stale tools on navigation, detach, disable, or debugger teardown
- retain already-reported execution contexts and bound discovery/OOPIF fan-out

## Why

The WebMCP protocol says enabling the domain reports all currently registered tools, but current Chromium enumerates only the inspected target's root frame. That leaves tools registered earlier in child frames invisible unless they register again dynamically.

- Protocol: https://chromedevtools.github.io/devtools-protocol/tot/WebMCP/
- Current Chromium implementation: https://chromium.googlesource.com/chromium/src/+/refs/heads/main/third_party/blink/renderer/core/inspector/inspector_web_mcp_agent.cc

## Validation

- real Chrome 151 unpacked-extension probe: parent, cross-site child, and nested cross-site grandchild tools were discovered and invoked; child navigation removed the complete child subtree
- `node test/run.js`: 1125 passed, 2 pre-existing baseline failures (changelog version mismatch and localized promotion assertion)
- `npm run test:security`: 60/60 passed
- `npm run test:ci`: 9 scenarios plus cloud capture passed
- merge-tree audit against current upstream `main`: no conflicts

### DIFF
--- a/src/chrome/src/cdp/cdp-client.js
+++ b/src/chrome/src/cdp/cdp-client.js
@@ -341,11 +341,13 @@ export class CDPClient {
     }
   }
 
-  _removeWebMCPTools(state, tools) {
+  _removeWebMCPTools(state, tools, sessionId = '') {
+    const owningSessionId = String(sessionId || '');
     for (const rawTool of Array.isArray(tools) ? tools : []) {
       const key = this._webMCPToolKey(rawTool?.frameId, rawTool?.name);
       const toolId = state.idsByKey.get(key);
       if (!toolId) continue;
+      if (state.toolsById.get(toolId)?.sessionId !== owningSessionId) continue;
       state.idsByKey.delete(key);
       state.toolsById.delete(toolId);
     }
@@ -536,7 +538,9 @@ export class CDPClient {
     register('WebMCP.toolsAdded', (params, source) => {
       this._storeWebMCPTools(state, params?.tools, source?.sessionId);
     });
-    register('WebMCP.toolsRemoved', params => this._removeWebMCPTools(state, params?.tools));
+    register('WebMCP.toolsRemoved', (params, source) => (
+      this._removeWebMCPTools(state, params?.tools, source?.sessionId)
+    ));
     register('WebMCP.toolResponded', (params, source) => (
       this._handleWebMCPResponse(state, params, source?.sessionId)
     ));
--- a/src/chrome/src/cdp/cdp-client.js
+++ b/src/chrome/src/cdp/cdp-client.js
@@ -273,6 +273,10 @@ export class CDPClient {
       nextToolId: 1,
       toolsById: new Map(),
       idsByKey: new Map(),
+      // Per tool-key generation bumped on store and remove. Discovery snapshots
+      // this map so a stale Runtime.evaluate result cannot resurrect or
+      // overwrite a concurrent registration change.
+      toolMutationGens: new Map(),
       pendingInvocations: new Map(),
       completedResponses: new Map(),
       childSessions: new Map(),
@@ -280,6 +284,16 @@ export class CDPClient {
       discoveryContextsUsed: 0,
       handlers: [],
     };
+  }
+
+  _webMCPToolMutationGen(state, key) {
+    return state.toolMutationGens.get(key) || 0;
+  }
+
+  _bumpWebMCPToolMutation(state, key) {
+    const next = this._webMCPToolMutationGen(state, key) + 1;
+    state.toolMutationGens.set(key, next);
+    return next;
   }
 
   _dropWebMCPSession(tabId, reason = 'WebMCP session closed') {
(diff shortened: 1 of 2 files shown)


src/chrome/src/cdp/cdp-client.js
```diff
@@ -297,13 +739,95 @@ export class CDPClient {
...
+          : ['', ...state.childSessions.keys()];
+        for (const affectedSessionId of affectedSessionIds) {
+          this._removeWebMCPSessionTools(state, affectedSessionId);
+          this._finishWebMCPSessionInvocations(
+            state,
+            affectedSessionId,
+            'The WebMCP document navigated before Chrome reported the invocation outcome.',
+          );
+        }
+        this._bumpWebMCPFrameMutation(state, frameId);
```
esokullu: **[bug]** On root `Page.frameNavigated`, only the navigated root `frameId` is frame-mutation-bumped (plus frames that already had tools, via `_removeWebMCPSessionTools`). Same-process descendant frames that never had a catalog entry are not invalidated. In-flight `_discoverExistingWebMCPFrameTools` captures `frameMutationGuard` before `Runtime.evaluate`; if a top-level navigation lands while evaluate is pending for a never-before-seen child frame, the stale snapshot still matches gen `0` and `_storeWebMCPTools` reinserts tools for the pre-navigation document. The per-frame navigation test covers the navigated frame itself, but not this root-nav / sibling-or-child frame case that the frame-gen machinery was introduced to solve.

**Suggestion:** On main-target root navigation (`!sessionId && !frame.parentId`), invalidate discovery more broadly: bump every known frame id (from `frameMutationGens`, `runtimeContexts`, and/or currently selected discovery contexts), or introduce a session-wide discovery epoch that merge checks in addition to per-key/per-frame gens. Add a regression test that starts discovery for a tool-less child frame, fires top-level `frameNavigated` mid-evaluate, and asserts the catalog stays empty.

src/chrome/src/cdp/cdp-client.js
```diff
@@ -297,13 +739,95 @@ export class CDPClient {
...
+        if (!sessionId) {
+          // Root document navigation starts a new discovery wave for later
+          // OOPIF re-attach cycles; without this the context budget stays spent.
+          state.discoveryContextsUsed = 0;
+          for (const nestedChild of state.childSessions.values()) nestedChild.frameIds.clear();
```
esokullu: **[bug]** Top-level navigation clears tools and child `frameIds`, but leaves `state.childSessions` populated and WebMCP/auto-attach still enabled on those OOPIF sessions. `_storeWebMCPTools` only rejects child tools when `!state.childSessions.has(owningSessionId)`, so a late `WebMCP.toolsAdded` (or a still-running child discovery pass) from a detaching pre-navigation OOPIF can repopulate the catalog after the embedding page has already navigated away. Permission hosts are still derived from that child session's frame tree when available, so the tool can look authorized for the old iframe origin on the new page.

**Suggestion:** On main-target root navigation, tear down child sessions the same way disable does: finish/cancel their invocations, `WebMCP.disable` + `setAutoAttach(false)` per child, delete them from `childSessions`, and rely on a fresh auto-attach wave for the new document. At minimum, remove children from `childSessions` immediately so store/discovery cannot re-enter, then async-teardown the CDP sessions.

src/chrome/src/cdp/cdp-client.js
```diff
@@ -249,35 +418,305 @@ export class CDPClient {
...
+          flatten: true,
+          filter: WEBMCP_IFRAME_TARGET_FILTER,
+        }],
+      ];
+      for (const [method, commandParams] of steps) {
+        if (!childStillOwned()) {
+          await this._teardownWebMCPChildSession(tabId, sessionId);
+          return;
+        }
+        await this.sendCommand(tabId, method, commandParams, sessionId).catch(() => {});
```
esokullu: **[suggestion]** Child-session setup runs `WebMCP.enable`, `Page.enable`, and `Target.setAutoAttach` with `.catch(() => {})` and continues into `_discoverExistingWebMCPFrameTools` even when `WebMCP.enable` failed. Discovery uses `document.modelContext.getTools()`, which can still return tools and insert them into the CDP-backed catalog with that `sessionId`. Later `WebMCP.invokeTool` / `cancelInvocation` on that session will fail, producing a listed tool that cannot be dispatched.

**Suggestion:** Track step success (or fail the child enable promise on `WebMCP.enable` errors). If enable fails, detach/teardown the child session and skip discovery instead of best-effort continuing.

src/chrome/src/cdp/cdp-client.js
```diff
@@ -297,13 +739,95 @@ export class CDPClient {
...
+      const child = state.childSessions.get(String(source?.sessionId || ''));
+      child?.frameIds.delete(frameId);
+    });
 
     state.enablingPromise = (async () => {
       try {
         await this.sendCommand(tabId, 'WebMCP.enable');
+        // Page domain is required for frameNavigated/frameDetached on the main
+        // target. Child OOPIF sessions enable Page separately when attached.
+        await this.sendCommand(tabId, 'Page.enable').catch(() => {});
```
esokullu: **[suggestion]** Main-target `Page.enable` is required for `Page.frameNavigated` / `Page.frameDetached` cleanup, but failures are swallowed. If Page domain enable fails while WebMCP enable succeeds, the session stays up with no frame lifecycle handlers effectively firing, so navigations will not clear tools or finish pending invocations.

**Suggestion:** Treat main-target `Page.enable` as required for a healthy WebMCP session (fail enable, or at least surface a structured warning and refuse to mark `enabled` without lifecycle coverage). Keep the child-path best-effort only where Page truly may be unavailable.

src/chrome/src/cdp/cdp-client.js
```diff
@@ -188,12 +274,74 @@ export class CDPClient {
       nextToolId: 1,
       toolsById: new Map(),
       idsByKey: new Map(),
+      // Per tool-key generation bumped on store and remove. Discovery snapshots
+      // this map so a stale Runtime.evaluate result cannot resurrect or
+      // overwrite a concurrent registration change.
+      toolMutationGens: new Map(),
```
esokullu, line 280: **[nit]** `toolMutationGens` and `frameMutationGens` grow for every tool key / frame id ever seen and are never pruned across long-lived WebMCP sessions with many navigations. Unlikely to matter at current caps, but it is unbounded state for an always-on debugger session.

**Suggestion:** When removing tools or handling root navigation, delete gens for removed keys/frames, or cap/prune maps similarly to `completedResponses`.

PR #443, opened 2026-07-21
feat: add guarded saved workflow replay

### LINKED ISSUE #442: Reusable workflows with guarded action caching and self-healing replay
## Problem

Repeated browser tasks currently pay the planning/tool-selection cost on every run, even after WebBrain has completed the same flow successfully. Custom skills provide Markdown instructions plus optional HTTP tools, but they are not an executable, parameterized workflow layer. Traces contain the successful tool chain, but replaying a trace verbatim would be unsafe and brittle:

- `ref_id` values only remain valid while their live DOM elements survive; navigation and SPA rerenders invalidate them.
- trace arguments may contain user-entered values, including credentials, so copying raw arguments into durable storage can persist secrets.
- a saved workflow must not inherit or bypass capability grants, submit confirmation, WebMCP confirmation, `/allow-api`, or completion verification.

This gap is also visible in `docs/claude-chrome-comparison.md`, which calls out missing `shortcuts_list` / `shortcuts_execute` primitives. Similar projects are converging on this layer:

- Stagehand caches observed actions/agent steps, reuses them without another LLM call, and falls back when the page no longer matches: https://docs.stagehand.dev/v3/best-practices/caching
- BrowserOS describes saved, repeatable visual workflows: https://docs.browseros.com/

## Proposed direction

Add a local-only **Saved Workflows** layer that can compile a successful run into a sanitized, parameterized workflow, replay matching steps deterministically, and hand control back to the agent when the page changes.

This should reuse the existing trace recorder as an input source, but define a separate workflow schema rather than treating `webbrain-trace/1` as executable.

### Workflow schema (suggested v1)

- name, description, created/updated timestamps, source WebBrain version
- allowed start origin and URL-family pattern
- declared runtime parameters, with sensitive parameters stored as placeholders only
- ordered steps containing:
  - tool name and sanitized/parameterized arguments
  - a semantic target descriptor where applicable (role, accessible name, label, field metadata, href, stable id/selector evidence)
  - expected precondition and postcondition
  - retry/fallback policy
- no reusable raw `ref_id`, element index, screen coordinate, password, OTP, API key, cookie, hidden form token, or captured request body/header

### Replay rules / safety invariants

1. Validate the current origin and URL family before starting.
2. Rebuild the current accessibility tree and resolve each target from its semantic descriptor. Never reuse a historical `ref_id`; ambiguity must fail closed.
3. Route replayed calls through the normal agent dispatch path so capability × origin permissions, submit confirmation, form verification, credential-field handling, WebMCP's mandatory per-call confirmation, and `/allow-api` still apply. Saving a workflow grants no authority.
4. Initially allow direct replay only for an explicit tool allowlist. Consequential actions require their normal fresh checks and an observable postcondition before advancing.
5. On locator mismatch, failed verification, unexpected navigation, or unknown action outcome, stop deterministic replay and resume the model from that step with current page evidence and the remaining workflow intent.
6. A self-healed locator/update is saved only after successful verification and explicit user opt-in; page content can never rewrite a workflow by itself.
7. Keep unattended/scheduled consequential execution out of the MVP. A future scheduler integration must preserve `scheduledRequireConsequentialConfirmation` semantics.

### Phase 1 — local saved workflow model and guided execution

- save/list/rename/delete/run UI
- local versioned schema + import/export validation
- parameter prompt that never pre-fills secret values from traces
- semantic target extraction and current-page re-resolution
- replay telemetry in traces (`matched`, `miss`, `model_fallback`, `healed`) without recording secret parameter values

### Phase 2 — guarded deterministic replay

- direct replay for a reviewed allowlist
- per-step pre/postcondition validation
- model fallback from the first mismatch instead of restarting the task
- optional, user-approved locator healing

### Phase 3 — product surface

- model-callable `workflows_list` and `workflow_execute` (or equivalent slash commands)
- optional scheduler integration
- visual graph editor only if the simpler list/editor proves useful

## Out of scope for the first implementation

- cloud sync or sharing workflows
- a credential vault
- replaying captured network bodies/headers or opaque API replay IDs
- unattended consequential workflows
- a visual graph builder

## Acceptance criteria

- [ ] Chrome and Firefox use the same workflow schema and pure normalization/validation tests.
- [ ] Saving a trace with form input does not persist secret literals or historical `ref_id` values.
- [ ] A workflow can rerun on an unchanged page with fewer LLM calls than the original run.
- [ ] A DOM rerender/navigation causes fresh target resolution rather than a stale action.
- [ ] Ambiguous targets, changed origins, permission denial, and unknown outcomes stop or fall back safely.
- [ ] Existing permission, submit-confirmation, `verify_form`, WebMCP, and completion invariants have regression coverage for replayed calls.
- [ ] Replay exposes hit/miss/fallback metrics so reliability and token savings can be measured.

## Why this fits WebBrain

Most of the necessary safety substrate already exists: local trace events with structured tool args/results, stable-within-document AX references plus semantic element metadata, deterministic capability/origin permissions, submit confirmation, verified field operations, loop detection, and scheduled-task policies. The missing piece is a durable compiler/executor between traces and the agent loop—not another set of site-specific prompt instructions.

alectimison-maker:
## Summary
- compile the latest successful traced run into a separate value-free saved-workflow schema
- add local list, save, run, and delete flows with ephemeral runtime parameter collection
- replay through fresh semantic target resolution and the existing permission, submit, form-validation, and verification gates
- fail closed on ambiguous targets, page-family drift, brittle selector-only actions, and unknown side-effect outcomes
- keep runtime values and fresh element references out of workflow storage, chat text, user memory, fallback prompts, UI telemetry, and replay traces
- preserve Chrome/Firefox parity and document the architecture, privacy, and security boundaries

## Testing
- node test/run.js: 1151 passed; 2 pre-existing main failures remain (changelog/package version mismatch and localized promotion expectation)
- node test/security/injection-corpus.mjs: 60/60 passed
- node scripts/build-zip.mjs: Chrome, Edge, and Firefox packages built successfully

Closes #442


src/chrome/src/agent/workflows.js
```diff
@@ -0,0 +1,707 @@
...
+    };
+  }
+  if (name === 'wait_for_element') {
+    const selector = cleanText(args.selector);
+    const text = cleanText(args.text);
+    const timeout = Number(args.timeout);
+    if (!selector && !text) return null;
+    return {
+      ...(selector ? { selector } : {}),
+      ...(text ? { text } : {}),
```
esokullu: **[bug]** `wait_for_element` still persists CSS `selector` arguments into the saved workflow. That contradicts the PR/docs fail-closed policy that discards brittle action CSS selectors (and the security model’s “no raw selectors” boundary). A selector like `.card:nth-child(2) button` can resolve to a different control after a page update, yet compile/normalize accept it without warning. `click` correctly rejects selector-only actions in the same file.

**Suggestion:** Treat selector-only `wait_for_element` like selector-only `click`: reject at compile/normalize, or require a text target / semantic postcondition and drop `selector` from storage. At minimum, refuse selectors that look structural (`nth-child`, long pure class chains) and document the exception if any selector retention remains intentional.

src/chrome/src/agent/agent.js
```diff
@@ -12177,6 +12187,199 @@ Rules: no prose intro, no conclusion, no "this screenshot shows...", no layout d
...
+          if (type === 'tool_result' && data?.name === step.tool) rawResult = data.result;
+          if (type === 'tool_call' && data?.name === step.tool) {
+            onUpdate(type, { ...data, args: redactWorkflowArgsForTelemetry(step.args, executionArgs), workflowReplay: true });
+            return;
+          }
+          if (type === 'tool_result' && data?.name === step.tool) {
+            onUpdate(type, { ...data, result: redactWorkflowResultForTelemetry(step.tool, data.result), workflowReplay: true });
+            return;
+          }
+          onUpdate(type, data);
```
esokullu: **[bug]** `replayUpdate` redacts only `tool_call` / `tool_result` for the active step. Every other update type is forwarded raw (`onUpdate(type, data)`), including `clarify` submit-confirmation payloads. Existing submit confirmation includes `summary` / `fields` / `changedFields` with live form values (see `_promptSubmitConfirmation` / form detection). Those values are runtime workflow parameters that were just typed into the page, and `compactRunUiData` does not strip `clarify` data before run-UI journal / session persistence. This breaks the security-model claim that “Replay telemetry and UI events redact runtime values.”

**Suggestion:** When `workflowReplay` is active, either (a) redact/omit field values from submit-confirmation and similar clarify payloads before `onUpdate`, or (b) pass a replay flag into the permission/submit helpers so they emit labels/hosts only. Mirror the same redaction in Firefox’s identical replay path.

src/chrome/src/agent/workflows.js
```diff
@@ -0,0 +1,707 @@
...
+    const url = safeHttpUrl(args.url);
+    if (!url) return null;
+    if (String(args.url || '').includes('?') || String(args.url || '').includes('#')) {
+      warnings.push('A navigation query or fragment was removed; provide dynamic destinations at run time.');
+    }
+    return { url };
+  }
+  if (name === 'go_back' || name === 'go_forward') return {};
+  if (name === 'click_ax') return target ? {} : null;
+  if (name === 'set_checked') {
```
esokullu, line 251: **[bug]** Semantic tools are compiled whenever `target` is any non-null object (`click_ax` returns `{}` if `target` is truthy; `type_ax` / `set_field` / `set_checked` similarly accept weak targets). Matching later requires `score >= 7` (`findWorkflowTarget`). A target with only `role`+`type` scores 5 and can never match; placeholder alone scores 5; many unlabeled controls will compile “successfully,” then always fall back or fail at run time with no save-time warning. That undermines the deterministic-replay product claim and can surprise users who thought `/workflow --save` captured a reusable flow.

**Suggestion:** At compile (and normalize) time, require `scoreWorkflowTarget(target, target) >= 7` (or an explicit identity field such as `id` / `fieldName` / `name` / `label` / `ariaLabel`). Otherwise skip the step with a clear warning, e.g. “Skipped click_ax: target is too weak to resolve uniquely.”

src/chrome/src/agent/workflows.js
```diff
@@ -0,0 +1,707 @@
...
+  const remaining = (workflow?.steps || []).slice(index).map((step) => ({
+    tool: step.tool,
+    target: step.target || null,
+    args: step.args || {},
+    expected: step.expected || null,
+  }));
+  const parameters = (workflow?.parameters || []).map((parameter) => ({
+    id: parameter.id,
+    label: parameter.label,
+    sensitive: parameter.sensitive === true,
```
esokullu, line 675: **[suggestion]** `workflowFallbackPrompt` embeds remaining step `args` as stored, including internal `{ "$workflowParam": "password" }` markers. Values are correctly omitted, but local/smaller models may try to type the marker object, treat it as literal text, or get stuck instead of re-asking the user. The parameter note helps, but the args JSON still looks like executable tool input.

**Suggestion:** When building the fallback prompt, rewrite parameter refs to an explicit placeholder string such as `"<ask user for parameter: password>"` and avoid emitting the internal `$workflowParam` object shape to the model.

src/chrome/src/agent/agent.js
```diff
@@ -12177,6 +12187,199 @@ Rules: no prose intro, no conclusion, no "this screenshot shows...", no layout d
...
+        estimatedLlmCallsSaved: matchedSteps,
+      });
+      onUpdate('tool_result', {
+        name: 'done',
+        result: { success: true, done: true, outcome: 'success', summary, workflowReplay: true },
+      });
+      traceStatus = 'done';
+      finalContent = summary;
+      return { status: 'completed', summary, matchedSteps, estimatedLlmCallsSaved: matchedSteps };
+    } finally {
```
esokullu, line 12385: **[suggestion]** `replaySavedWorkflow`’s `finally` clears running-tab / loop / completion state, but not several fields that normal `processMessage` clears (`_clickAxCdpFallbacks`, cost state, plan guards, active skills reset, etc.). A workflow step that exercises click-AX CDP fallback or other per-tab run state can leave residue that affects the next Act turn or the immediate fallback `processMessage` call.

**Suggestion:** Align cleanup with `processMessage`’s `finally` (or extract a shared `_endRunCleanup(tabId, token)` used by both paths), at least clearing `_clickAxCdpFallbacks` and cost/plan run maps.

src/chrome/src/ui/sidepanel.js
```diff
@@ -2965,6 +2977,191 @@ async function forgetUserMemory(id, tabId = currentTabId) {
   }
 }
 
+const SAVED_WORKFLOW_FAILURE_REASON_KEYS = {
```
esokullu, line 2980: **[suggestion]** `SAVED_WORKFLOW_FAILURE_REASON_KEYS` only maps `conversation_required`, `no_successful_trace`, `no_replayable_steps`, and `not_found`. Compiler reasons such as `name_required`, `successful_run_required`, `http_start_url_required`, and `normalization_failed` surface as raw `Workflow error: name_required`-style strings. Empty `/workflow --save` is a common case.

**Suggestion:** Map the remaining compiler/store reasons to localized strings (or reuse the existing no-trace / no-steps messages where appropriate) in both Chrome and Firefox sidepanels.

src/chrome/src/agent/workflows.js
```diff
@@ -0,0 +1,707 @@
...
+  };
+  addExact('id', 12);
+  addExact('fieldName', 9);
+  addExact('label', 8);
+  addExact('ariaLabel', 8);
+  addExact('name', 7);
+  addExact('href', 7);
+  addExact('placeholder', 5);
+  addExact('selector', 5);
+  addExact('type', 3);
```
esokullu, line 707: **[nit]** `scoreWorkflowTarget` awards points for `selector`, but `TARGET_FIELDS` / `normalizeTarget` never keep `selector`, so this branch is dead. Harmless, but suggests leftover design from an earlier target shape.

**Suggestion:** Remove the `selector` scoring branch (or stop advertising selector as target evidence) to keep the matching model honest.

esokullu (changes requested):

esokullu:
this is great, is this the phase one of what you describe at https://github.com/webbrain-one/webbrain/issues/442

PR #447, opened 2026-07-22
feat: add conditional /watch polling and beep alerts

### LINKED ISSUE #410: Add `/watch` shorthand with conditional `/beep` notifications
## Summary

Add `/watch` as an opinionated shortcut over the existing scheduled-task system. It should poll the current page every minute and optionally alert the user only when the watched condition occurs.

## Proposed UX

```text
/watch [--keep] <condition and action> [/beep]
```

Example:

```text
/watch if a new commit appears, post "@codex pls review" /beep
```

## Behavior

- Run in Act mode against the current page URL.
- Perform the first check immediately, then poll every minute.
- Default behavior: stop after the first successful trigger.
- `--keep`: continue watching for subsequent distinct events.
- For relative conditions such as “new commit,” the first run establishes a baseline.
- For absolute conditions such as “when CI is green,” the first run may trigger immediately.

## Scheduler contract

Store watches as normal scheduled tasks with watch-specific metadata rather than building a second scheduler:

```js
source: "watch"
watch: {
  keep: boolean,
  beep: boolean
}
```

Each run receives the previous observation and last triggered event key.

- `done(partial)`: no matching change; schedule the next poll.
- `done(success)`: condition and requested action were verified; stop unless `--keep` is active.
- `done(failed)` or runtime error: fail the watch instead of retrying every minute indefinitely.

## Conditional `beep`

Expose a scoped tool only while running a watch created with the `/beep` modifier:

```js
beep({
  event_key: string,
  message?: string
})
```

- Do not add `beep` to ordinary agent runs.
- Arm the alert during the run and play it only if the run finishes successfully.
- Suppress repeated alerts for the same `event_key`.
- Keep ordinary polling runs silent.
- Honor the existing `notifySound` setting.
- Deliver the sound from the background so it also works when the side panel is closed.
- Reuse Chrome’s shared offscreen host and Firefox’s background page.
- Prevent the side panel from playing the same alert twice.

## UI and compatibility

- Add `/watch` and `--keep` to autocomplete, `/help`, localized strings, and both Chrome and Firefox side panels.
- Show watch status, interval, stop-on-trigger/keep-watching mode, and last observation in the scheduled-job card.
- Keep existing `/schedule`, recurring jobs, and completion-sound behavior unchanged.

## Acceptance criteria

- `/watch`, `--keep`, and trailing `/beep` parse consistently across Chrome and Firefox.
- Empty prompts and invalid option combinations are rejected locally.
- No-change polls remain silent and continue.
- Successful one-shot watches alert once and stop.
- `--keep` watches alert once per distinct event.
- Failed watches stop without producing an alert.
- Alerts work with the side panel closed and respect `notifySound`.
- Existing scheduler and completion-sound regression tests remain green.
- `node test/run.js` and `npm test` pass.

## Reference

Chrome offscreen API: https://developer.chrome.com/docs/extensions/reference/api/offscreen

alectimison-maker:
## Summary

- add `/watch [--keep] [--secs <30-120>] [--long | --short] <condition and action> [/beep]` for immediate, page-bound conditional polling
- extend the durable scheduler with watch metadata, baseline-aware `partial` polls, one-shot/keep outcome handling, safe untrusted observation replay, and duplicate coalescing that remains stable after the first poll
- run watches in dedicated inactive helper tabs; a diverged helper is replaced instead of navigating the initiating/current tab back to the watched URL
- reuse one sticky assistant message per watch and replace it with the latest poll result, avoiding blank bubbles and stale observations
- expose a scoped `beep({event_key, message?})` tool only to `/beep` watches; persist and dedupe event keys before background audio, and play only after verified `done(outcome="success")`
- support distinct short/default/long tones through Chrome MV3 offscreen audio and the Firefox background page, honoring `notifySound` without a second side-panel chime
- surface watch interval/mode/alert/last-observation/warning state in scheduled cards, localize command/status copy across all shipped locales, and document behavior/security boundaries

## Safety properties

- prior page-derived observations are nonce-delimited and boundary markup is stripped before the next poll
- ordinary agent runs cannot see or invoke `beep`; one poll can arm only one fresh event key
- duplicate, partial, failed, and missing-outcome runs do not play an alert
- a verified successful action that omits the optional `beep` arm completes or continues without sound and records a warning; it does not destroy a `--keep` watch
- a fresh event key is persisted before best-effort audio playback
- watch polling never forces a user-controlled tab back to the watched URL

## Tests

- `node test/run.js`: **1269 passed**, with 1 pre-existing `main` failure (CHANGELOG `25.8.0` vs package `25.8.5`)
- `npm run test:security`: **60/60 passed**
- `npm run test:fixtures`: **125/125 passed**
- syntax checks pass for every modified JavaScript file; `git diff --check` passes

Closes #410


src/chrome/src/agent/scheduler.js
```diff
@@ -766,6 +852,65 @@ export class ScheduledJobManager {
...
+        beepStyle: parsed.beepStyle,
+        intervalSeconds: parsed.intervalSeconds,
+        baselineEstablished: false,
+        lastObservation: null,
+        lastTriggeredEventKey: null,
+        lastTriggeredAt: null,
+      },
+      scheduledAt: createdAt,
+      nextRunAt: iso(this.now() + 1000),
+      immediate: true,
```
esokullu, line 909: **[bug]** Watch jobs are created with `immediate: true` and `nextRunAt ≈ now+1s`, then `_completeWatch` sets `immediate: false` and advances `nextRunAt` by the watch interval. `sameScheduledIntent` still requires matching `scheduledJobIsImmediate` and near scheduled times, so after the first poll an identical `/watch` is no longer treated as a duplicate live job. Users can stack many concurrent Act-mode watches for the same page/condition (amplified by 30–120s intervals and LLM cost).

**Suggestion:** For `source === 'watch'`, dedupe live jobs by target + watch payload only (ignore `immediate` / near-time). Extend the existing “identical live watch should be deduped” test to re-create after one successful partial poll.

src/chrome/src/ui/sidepanel.js
```diff
@@ -2515,6 +2546,9 @@ async function handleScheduledJobEvent(data, tabId) {
   } else if (event === 'completed') {
     ensureScheduledTerminalMessage(job);
     await settleScheduledRun(event, job, runTabId);
+  } else if (event === 'polled' || event === 'triggered') {
```
esokullu, line 2591: **[bug]** Keep watches emit `running` on every poll, and the existing handler always does `addMessage('assistant', '')` with the same `dataset.scheduledJobId`. The new `polled`/`triggered` path only settles the first matching assistant node, and only fills text when that node is still empty. Result: each 30–120s poll leaves an additional empty assistant bubble in chat, while later observations never replace the first filled one.

**Suggestion:** For `job.source === 'watch'` on `running`, reuse `findScheduledAssistantMessageForJob(jobId)` (or a single sticky run row) instead of creating a new message; on `polled`/`triggered`, update that message’s text to the latest `lastResult` rather than gating on `!textEl.textContent.trim()`.

src/chrome/src/agent/scheduler.js
```diff
@@ -766,6 +852,65 @@ export class ScheduledJobManager {
...
+      prompt: parsed.prompt,
+      schedule: {
+        type: 'recurring',
+        run_at: createdAt,
+        interval_seconds: parsed.intervalSeconds,
+      },
+      target: {
+        type: 'url',
+        url: parsed.url,
+        ...(tabId != null ? { tabId } : {}),
```
esokullu: **[bug]** `createWatchJob` stores the initiating `tabId` on a `type: 'url'` target. Existing `_resolveTab` will `tabs.update` that same tab back to the watched URL whenever the tab has navigated away. At 30–120s this repeatedly hijacks the user’s active browsing session (and can wipe SPA state) instead of following the page or using a background tab.

**Suggestion:** Prefer a dedicated inactive tab for watches: omit initiating `tabId` from the URL target (always `tabs.create({ active: false })`), or if URL diverges create/reuse a background tab rather than forcing `tabs.update` on the original tab. Document the chosen binding semantics in README/architecture.

src/chrome/src/agent/scheduler.js
```diff
@@ -1084,10 +1229,158 @@ export class ScheduledJobManager {
...
+      && watchAlert.duplicate === true
+      && !!eventKey
+      && eventKey === job.watch?.lastTriggeredEventKey;
+    const freshAlert = lastOutcome === 'success'
+      && job.watch?.beep === true
+      && watchAlert.armed === true
+      && !!eventKey
+      && !duplicateAlert
+      && eventKey !== job.watch?.lastTriggeredEventKey;
+    const alertContractFailed = lastOutcome === 'success'
```
esokullu: **[suggestion]** For `/beep` watches, any `done(outcome="success")` without a fresh armed `event_key` permanently fails the entire job (`alertContractFailed`). That is intentionally fail-closed for audio, but local/small models frequently omit optional tools; a single miss destroys a long-running `--keep` watch after it may already have performed the user-requested action.

**Suggestion:** Consider treating missing arm as a non-fatal “success without alert” (complete/continue without sound, record a warning), or only fail closed when beep was armed with a key that fails the contract. Keep the stronger fail-closed path if product intent is strict, but call it out in the slash help / created message.

src/chrome/src/ui/sidepanel.js
```diff
@@ -2464,7 +2488,10 @@ async function settleScheduledRun(event, job, tabId = currentTabId) {
   if (assistantEl) {
```
esokullu, line 2522: **[suggestion]** Settlement still only injects `job.lastResult` when the assistant text is empty. Even after Issue 2 is fixed to reuse one bubble, keep-watch users will not see refreshed observations in chat unless this condition is relaxed for `polled`/`triggered`.

**Suggestion:** For watch poll events, always replace/update the message text with the latest observation (still via `formatMarkdown`), and keep the empty-only gate for true terminal schedule completions if desired.

src/chrome/src/ui/locales/en.js
```diff
@@ -297,6 +298,13 @@ export default {
   "sp.slash.record_transcribe": "Record and save a Whisper transcript after stop",
```
esokullu, line 319: **[nit]** `/watch` is `outOfBand: true` and works while busy, but `sp.slash.busy_only_oob` still lists only the older OOB commands and does not mention `/watch`.

**Suggestion:** Add `/watch` to the busy-only OOB toast string in all locales (Chrome and Firefox).

esokullu (changes requested):

alectimison-maker:
Addressed all six review points in `70b8b4e8` after merging the latest `main`:

- watch dedupe now ignores drifting immediate/time fields and is tested after a real partial poll
- each watch reuses one sticky assistant message and replaces it with the latest observation
- watches no longer bind or navigate the initiating tab; dedicated helpers stay inactive and diverged helpers are replaced
- a successful run that misses the optional beep arm now completes/continues without sound and records a warning
- `/watch` is included in the busy OOB message for every Chrome/Firefox locale (including locales added since the original PR)
- README and architecture docs now describe helper-tab and missed-alert semantics

Validation: `node test/run.js` 1269/1270 (only the current-main CHANGELOG 25.8.0 vs package 25.8.5 mismatch), security 60/60, fixtures 125/125. The GitHub smoke/WebMCP E2E check is green. The remaining Vercel status is the team authorization check, not a code/test failure.

esokullu:
# Review — `/watch` conditional polling and beep alerts

The design is well thought out: the arm-then-verify-then-persist-then-play ordering, fail-closed `beep` gating (tool list filter *and* an `executeTool` policy re-check), event-key dedupe surviving restarts, and the decision to never navigate a user-visible tab back to the watched URL are all correct and tested. Chrome/Firefox parity was verified mechanically — the only divergence mirrors a pre-existing platform difference (Firefox activates scheduled-task tabs; watch tabs correctly stay inactive on both).

## Issues

**1. Helper-tab leak when the watched URL diverges every poll (main concern).**
`sameTargetUrl` is exact `URL.href` equality, and for watch jobs the "navigate the helper back" fallback is skipped, so a diverged helper always falls through to `tabs.create`. Nothing ever calls `tabs.remove` — the old helper is abandoned, not closed. Many real pages diverge on *every* load: canonical redirects, appended query/tracking params, SPAs rewriting the URL via `history.pushState`, consent/login interstitials. For such a page, a `--keep --secs 30` watch abandons a new background tab every 30 seconds — 120 leaked tabs/hour. Fix: (a) `tabs.remove` the diverged helper before creating its replacement (it's a watch-owned inactive tab, so closing it is safe), and (b) close the helper on terminal states (`completed`/`failed`/cancel).

**2. One transient failure permanently kills a `--keep` watch.**
`_completeWatch` marks the job `failed` on any `failed` or missing outcome — including a flaky network fetch, a temporarily unavailable page, or an LLM provider error mid-poll. For a watch the user expects to run for hours, a single hiccup at 3 a.m. silently ends it. Suggest tolerating N consecutive failures before stopping.

**3. Deduped creation reports success as if new.**
`createWatchJob` returns `{ deduped: true, ... }`, but the sidepanel branch treats any `scheduledAt` as success and shows `sp.watch.created` ("Started watching…"). Telling the user the page is already being watched would be more honest.

**4. Post-duplicate arming loophole (minor).**
After `beep` returns `duplicate: true` (which does not set `armedEventKey`), the same poll can call `beep` again with a *different* key and arm it; the scheduler's `onUpdate` keeps the last qualifying result, so that fresh arm wins. The "one fresh key per poll" invariant holds, but "on duplicate, finish with `partial`" is enforced only by prompt. Cheap hardening: have a `duplicate` response also lock `armedEventKey` so a subsequent fresh arm in the same poll is rejected.

## Minor notes

- `createWatchJob({ tabId = null, ... })` never uses `tabId` — dead parameter.
- `wrapWatchObservation` uses `Math.random` for the nonce. Fine in practice because boundary markup is stripped first, but `crypto.getRandomValues` is equally cheap.
- The closing delimiter `</untrusted_page_content id="...">` puts an attribute on a closing tag — invalid as markup; deliberate as a prompt delimiter, but deserves a comment.
- The watch card reuses the `sp.slash.watch_keep` *option description* string as a status label; a dedicated status key would read better in some locales.
- A watch **without** `/beep` completes with no sound at all (watch completions are excluded from `notifyCompletion`). Matches the documented "alerts only via /beep" model, but worth confirming this is intended UX.

## Test coverage, conventions, security

- **Tests**: excellent — parser parity across browsers, interval/target/style validation, beep scoping and arm/dedupe semantics, alert-once-per-key ordering including the persist-before-play policy snapshot, baseline establishment with injection-attempt content, helper-tab divergence, dedupe stability after the first poll, and the missed-optional-beep warning path.
- **Conventions**: mirrored Chrome/Firefox trees, locale key updates across all 21 locales in both builds, and docs all follow existing project patterns.
- **Security**: the claimed properties check out in the code — `beep` is invisible and denied outside `/beep` watches, boundary markup is stripped before nonce-wrapping, event keys persist before best-effort playback, offscreen audio receives only a style string, and `notifySound === false` mutes both platforms.

## Verdict

Solid, carefully engineered feature with unusually good test discipline. Follow-up commits addressing the issues above are being pushed to this branch directly.

esokullu:
Pushed 57f34bbc addressing everything from the review above:

- **Helper-tab leak (issue 1):** a diverged helper is now closed before its replacement is created, and the helper is also closed when the watch completes, fails, is cancelled, or is deleted. Covered by new assertions in the divergence, one-shot, failure, and cancel paths.
- **Transient failures (issue 2):** failed polls, missing `done` outcomes, and thrown runs are tolerated up to 3 consecutive occurrences (`MAX_WATCH_CONSECUTIVE_FAILURES`), tracked in `watch.consecutiveFailures` and reset on any successful/partial poll. A failed poll never overwrites the baseline observation. `_markFailed` routes watch jobs through the same tolerance, so a provider error mid-poll no longer kills a `--keep` watch.
- **Dedupe message (issue 3):** `/watch` on an already-watched page now shows a new `sp.watch.exists` string (added to all 21 locales in both builds) instead of claiming a new watch started.
- **Post-duplicate arming (issue 4):** a `duplicate: true` beep response now also locks the run's event key, so the same poll can no longer arm a different fresh key afterwards.
- **Minors:** `crypto.getRandomValues` nonce with a comment explaining the closing-delimiter id; dead `tabId` parameter removed from `createWatchJob`; keep-mode card label got its own `sp.scheduled.watch_keep` status string; README + architecture docs updated.

Tests: `node test/run.js` 1269 passed (same pre-existing CHANGELOG-version failure as main), `test:security` 60/60, `test:fixtures` 125/125.

PR #2608, opened 2026-07-27
fix(firefox): implement documented global keyboard shortcuts

Ayush7614:
## Summary

Firefox `/help` advertises keyboard shortcuts (`Ctrl/Cmd+/`, mode switches, Escape abort), but `sidepanel.js` never registered a global keydown handler — so the shortcuts did nothing.

This ports Chrome's `handleGlobalKeydown` to Firefox (without Chrome-only recording Escape handling).

## Why

Real Firefox parity gap: documented UX that was broken. Not tied to any open issue.

## Test plan

- [x] `node test/run.js`: **1350 passed, 1 failed** (existing changelog test on `main`)
- [x] Firefox sidepanel: `Ctrl/Cmd+/` focuses input
- [x] `Ctrl/Cmd+Shift+A/X/D` switch Ask/Act/Dev when idle
- [x] Escape aborts an active run; Escape with slash menu open only dismisses the menu

### DIFF
--- a/src/firefox/src/ui/sidepanel.js
+++ b/src/firefox/src/ui/sidepanel.js
@@ -9319,8 +9319,16 @@ async function handleGlobalKeydown(e) {
   const isOtherFormField = e.target !== inputEl && (tag === 'INPUT' || tag === 'TEXTAREA' || tag === 'SELECT');
 
   if (e.key === 'Escape') {
+    // IME Escape cancels a composition candidate — never abort the run for that.
+    if (e.isComposing) return;
     const slashMenuOpen = !!slashCommandMenuEl && !slashCommandMenuEl.classList.contains('hidden');
     if (slashMenuOpen) return;
+    // Provider/language pickers close on Escape in bubble/target handlers; do not
+    // abort the active run while those listboxes are open.
+    const providerPickerOpen = !!providerPickerMenu && !providerPickerMenu.classList.contains('hidden');
+    if (providerPickerOpen) return;
+    const languagePickerOpen = !!languagePickerMenu && !languagePickerMenu.classList.contains('hidden');
+    if (languagePickerOpen) return;
     if (isProcessing) {
       e.preventDefault();
       abortRun();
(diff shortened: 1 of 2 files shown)


esokullu:
The shortcuts work in their basic cases, but the capture-phase handler steals native international text-entry events and can abort active runs while users dismiss other open controls. These regressions make the patch unsafe to merge as written.

Full review comments:

- [P1] Preserve AltGr slash input in the composer — `/Users/barack/Documents/webbrain-webbrain-one-3/src/firefox/src/ui/sidepanel.js:9334-9337`
  On layouts where `/` is produced with an AltGr or Ctrl+Alt chord and Firefox exposes `ctrlKey` and `altKey`, this condition treats ordinary text entry as Ctrl+/. Because the main composer is exempt from `isOtherFormField`, it prevents insertion and moves the caret instead, making `/` and slash commands impossible to type; exclude Alt/AltGraph-modified events.

- [P1] Ignore Escape while an IME composition is active — `/Users/barack/Documents/webbrain-webbrain-one-3/src/firefox/src/ui/sidepanel.js:9321-9326`
  While an active run is in progress, the composer remains available for queued messages. If an IME dispatches Escape to cancel a candidate or composition (`e.isComposing === true`), this capture listener runs before the textarea can consume it and calls `abortRun()`, turning normal CJK text editing into cancellation of the agent run.

- [P1] Let open pickers consume Escape before aborting — `/Users/barack/Documents/webbrain-webbrain-one-3/src/firefox/src/ui/sidepanel.js:9322-9326`
  When an active run is in progress and the provider or language listbox is open, Escape first reaches this capture handler before the picker's target/bubble handlers. Since only slash autocomplete is checked, `abortRun()` executes and then the picker closes, so a routine listbox dismissal unexpectedly cancels the task; open pickers need the same precedence as the slash menu.

Ayush7614:
Addressed the three P1s from review in a follow-up commit:

1. **AltGr slash** — ignore `altKey` / `AltGraph` before treating Ctrl/Cmd chords as shortcuts, so international `/` entry is not stolen.
2. **IME Escape** — return early when `e.isComposing` so composition cancel does not call `abortRun()`.
3. **Open pickers** — give provider and language listboxes the same Escape precedence as slash autocomplete (do not abort while they are open).

Tests updated to assert the three guards precede `abortRun()`.

PR #2638, opened 2026-08-02
agent: add adapter for JD

alectimison-maker:
## Summary

Adds a mirrored JD.com adapter for Chrome and Firefox.

Without this adapter, anonymous JD product navigation can land on JD's passport login flow without recovery guidance, and shopping actions can blur the distinction between adding to cart, reviewing checkout, and creating an order.

- Scope matching to JD's consumer shopping, checkout, order, and login hosts.
- Handle `passport.jd.com/new/login.aspx` redirects without navigation loops.
- Distinguish self-operated and third-party offers, variant and delivery-dependent totals, cart review, checkout review, and final order submission.
- Require explicit user confirmation before `提交订单` and verify `订单编号` before reporting success.
- Add positive, negative, phishing-lookalike, guidance, and Chrome/Firefox parity assertions.

## Validation

- Targeted JD production assertions: passed.
- `node test/run.js`: 1396/1397 passed; the single failure predates this branch and is the repository's `CHANGELOG.md` latest version (`26.0.0`) not matching `package.json` (`26.0.3`).
- `node test/security/injection-corpus.mjs`: 60/60 passed.
- Anonymous Chrome/Playwright read-only check: JD search, product/login redirect, and cart final URLs all selected the `jd` adapter; the product URL redirected to JD's `passport.jd.com` login with `czLogin=1`.
- `git diff --check`: passed.

### DIFF
--- a/src/chrome/src/agent/adapters.js
+++ b/src/chrome/src/agent/adapters.js
@@ -16337,6 +16337,21 @@ const ADAPTERS = [
 - "ポイント", SPU, and スーパーDEAL are conditional rewards, not an immediate cash-price reduction; campaigns may require entry, membership, payment methods, or have caps, and shipping/tax do not earn ordinary points. Report the payable total separately from expected points.`,
   },
 
+  // ─── Regional — JD.com (China) ───────────────────────────────────────
+  {
+    name: 'jd',
+    category: 'general',
+    matches: (url) => /^https?:\/\/(?:(?:www|search|item|cart|trade|order|passport)\.)?jd\.com\//.test(url),
+    notes: `
+- Treat www.jd.com, search.jd.com, item.jd.com, cart.jd.com, trade.jd.com, order.jd.com, and passport.jd.com as JD's desktop shopping flow as of 2026-08. Expect anonymous visits to search or product pages to redirect to passport.jd.com/new/login.aspx with czLogin=1.
+- When that login redirect appears, STOP and ask the user to complete QR code, password, or SMS login manually. Do not loop back to the product/search URL or claim it is unavailable. After login, continue the encoded ReturnUrl and re-read the destination page.
+- On product pages, "搜全站" searches all of JD while "搜本店" restricts results to the current store. Use "搜全站" unless the user explicitly wants more products from that store.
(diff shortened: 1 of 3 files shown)


esokullu:
Thanks — this is a well-shaped adapter. It follows `CONTRIBUTING.md` closely: correct four-field shape, placed in the Regional block next to Rakuten/Allegro, dated, imperative voice, 7 bullets against the 5–15 budget, real on-page strings rather than vague advice, and it names both an anti-pattern (`立即购买` skips the cart) and a success indicator (`订单编号`). The mutation boundary — cart → `去结算` → explicit confirmation before `提交订单` — is exactly the posture we want.

Verified locally: `node test/run.js` gives 1396 passed / 1 failed on this branch, and the one failure is the `CHANGELOG.md` vs `package.json` version check that also fails on `main` (1395/1 there). Your validation note was accurate. I also probed the matcher directly — `item.jd.com.evil.com`, `item.jd.com@evil.com`, `evil.com@item.jd.com` and `xitem.jd.com` are all correctly rejected.

I pushed one follow-up commit rather than sending it back for a round trip:

- **Broadened host coverage.** The enumerated list was missing surfaces the notes themselves assume. `mall.jd.com` is where `搜本店` actually lives, so bullet 3 was describing a page the adapter didn't select on; `list.jd.com` (category browsing) is a primary comparison surface; and the `m.jd.com` mobile hosts were uncovered (Coupang already includes `m.`). Matcher is now `(?:(?:www|search|so|list|item|mall|cart|trade|order|passport)\.)?(?:m\.)?jd\.com\/`, which keeps `corporate.jd.com` and `help.jd.com` excluded as your test intends.
- **Softened the login-redirect claim.** "Expect anonymous visits ... to redirect" overstates it — anonymous JD browsing frequently works and the redirect is intermittent, so asserting it risks the agent preemptively stopping or reporting a login wall it never hit. Now "may redirect"; bullet 2 already handled the conditional case correctly.
- **Clarified the ReturnUrl step** — JD returns there itself after login, so the instruction is to re-read the destination rather than re-navigate.
- **Added `APP专享价`** alongside `京东价` as another conditional price.
- **Extended the test** to the new hosts plus a userinfo lookalike (`item.jd.com@phishing.example`), which locks in the trailing-`/` behavior a future host-list edit could otherwise break.

`jd` notes stay byte-identical across the Chrome and Firefox builds, and your parity assertion still covers that. Merging.

PR #2640, opened 2026-08-02
agent: add adapter for 12306

alectimison-maker:
## Summary

Adds a mirrored China Railway 12306 adapter for Chrome and Firefox.

Without this adapter, the agent can confuse live availability, a paid waitlist, a queued order, and a successfully booked ticket, or retry around identity verification.

- Scope matching to the official 12306 home, ticketing, order, waitlist, and passport hosts.
- Preserve exact stations, dates, trains, seat classes, passengers, and ticket types through the booking flow.
- Distinguish `有`, numeric inventory, `无`, `预订`, and `候补` states.
- Stop for QR/app/SMS/identity checks and require explicit confirmation before order, waitlist, or payment actions.
- Verify success from the final `订单状态`, including distinct waitlist outcomes.
- Add positive, negative, phishing-lookalike, label, transaction-safety, and Chrome/Firefox parity assertions.

## Validation

- Targeted 12306 production and guidance assertions: passed.
- Anonymous Chrome/Playwright read-only check: official home, ticket-search, and waitlist-help pages returned HTTP 200 and every final URL selected the `railway-12306` adapter.
- `node test/run.js`: 1397/1398 passed; the single failure predates this branch and is the repository's changelog/package version mismatch.
- `git diff --check`: passed.

## Compatibility and risks

The change is prompt-only and mirrored across both browser builds. No booking, passenger, identity, waitlist, or payment mutation was performed; those authenticated steps are grounded in the official 12306 flow and explicitly fail closed for user confirmation.

### DIFF
--- a/src/chrome/src/agent/adapters.js
+++ b/src/chrome/src/agent/adapters.js
@@ -16234,6 +16234,20 @@ const ADAPTERS = [
 - Price alerts require sign-in; "Hopper" predictions on price trends are advisory, not guarantees.`,
   },
   {
+    name: 'railway-12306',
+    category: 'general',
+    matches: (url) => /^https?:\/\/(?:(?:www|kyfw|passport)\.)?12306\.cn\//.test(url),
+    notes: `
+- Treat 12306.cn, www.12306.cn, kyfw.12306.cn, and passport.12306.cn as China Railway's official desktop flow as of 2026-08. Start from the ticket form's "出发地", "到达地", and "出发日期" controls; choose the exact station when a city has multiple stations and re-read both endpoints after using the swap control.
+- Read each result row by train number and seat class. "有" means seats are currently available, a number is the remaining count, "无" means sold out, and "候补" is a waitlist action; do not treat a displayed row or estimated availability as a reservation.
+- Stop for the user when "账号登录", "扫码登录", "人证核验", or "手机验证" appears. Do not retry around QR, app, SMS, or identity checks, and do not add or edit a passenger's real-name identity data unless the user explicitly asks.
+- Use "预订" only for an available train and confirm the exact date, train, departure/arrival stations and times, seat class, passenger, and ticket type on the passenger-confirmation page. Similar station names and overnight arrival dates are not interchangeable.
+- Treat "候补" as a paid waitlist, not a booked ticket. Read every requested train/seat combination, whether "接受新增列车" or "接受无座" is enabled, and the "截止兑现时间" before continuing; the "预付款" can use the highest requested fare and successful fulfillment can create a paid order automatically.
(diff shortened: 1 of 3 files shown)


esokullu:
The 候补 modeling here is the strongest thing in any of your three adapters. Paid waitlist ≠ ticket, `预付款` charging at the *highest* requested fare, fulfillment auto-creating a paid order, `截止兑现时间`, and the four terminal states split out — an agent gets every one of those wrong by default, and the money consequences are real. "A queue dialog, payment redirect, or submitted form is not proof of a ticket" is an unusually precise success criterion.

Verified: 1397/1398 on your branch, and merged against current `main` (which now has the JD and Taobao adapters) it auto-merges clean at 1399/1400 — the one failure being the pre-existing changelog check. `injection-corpus.mjs` 60/60. Matcher probe rejects `12306.cn.evil.com`, `kyfw.12306.cn@evil.com`, `evil.com@kyfw.12306.cn`, `x12306.cn`, `95306.cn`.

One blocking issue, which I fixed and pushed:

**`epay.12306.cn` was unmatched.** Bullets 6 and 8 are specifically about payment — don't click a payment control without explicit confirmation, keep payment on official 12306 domains — but the matcher stopped at `www`/`kyfw`/`passport`. If the flow hands off to `epay.`, the adapter stops firing at exactly the step where money moves, and the agent's first turn there looks unguarded with nothing signalling that the rules lapsed. That's worse than not writing the payment guidance at all. DNS-confirmed and now matched: `epay`, `mobile`, `cx`, `dynamic`, `travel` (`bxfw` and `zxsc` don't resolve, so I left them out).

Also pushed:

- **A verification/anti-bot interstitial bullet**, on the mercado-libre STOP pattern. Deliberately written *without* a Chinese literal — see the questions below.
- **Tightened assertions.** `/有.*无.*候补/s` and `/订单.*状态/s` pass on nearly any Chinese text containing those characters; they'd survive a real regression. Anchored to phrases, and added userinfo + `epay` lookalike negatives.

**Three things only you can settle** — I could not verify them and did not guess:

1. Is `epay.12306.cn` actually the payment step in the current flow? I confirmed it *resolves*, not what it does. I hedged the note to "a step can hand off between 12306 hosts (for example kyfw to epay)" rather than asserting it.
2. Does payment ever leave `*.12306.cn` (e.g. to a bank or Alipay host)? If so, bullet 8's containment rule needs a carve-out, because as written the agent will treat a legitimate bank redirect as suspicious.
3. What does the verification wall actually look like — is there a stable visible string worth naming? A concrete literal beats my generic description, but I'd rather have no string than a wrong one.

Separately, worth knowing for all three of your adapters: Chrome's auto-translate kicked in on me mid-session on taobao.com and rewrote `搜索` → "search" in the DOM. If it was on during any of your label checks, the Chinese literals you read may have been translated text rather than the real strings.

Merging — the adapter fails closed, so if any of its claims are wrong the agent stops and asks rather than buying a ticket.

PR #2683, opened 2026-08-04
fix(agent): skip invalid JSON candidates

alectimison-maker:
## Problem

Planner and classifier responses can contain an illustrative or malformed object before the intended JSON result. The extractor stopped after either a balanced candidate that failed `JSON.parse` or an opening brace that never closed, so a later valid result was discarded.

## Change

- continue scanning after a balanced candidate fails `JSON.parse`
- resynchronize after an unterminated opening brace, capped at 16 restarts
- preserve fenced-output priority, string escaping, and first-valid-object behavior
- mirror the implementation across Chrome and Firefox
- add malformed-candidate, unbalanced-prefix, bounded-work, and planner regressions

## Scope and consumers

The extractor serves the planner, progress-intent, safety-refusal, plan-only, and visible-media location paths. Each consumer still applies its existing shape checks; this change only makes a later parseable JSON value reachable. An earlier valid but semantically unrelated object still wins and can cause the consumer to return `null`; choosing the first value that normalizes is intentionally out of scope.

## Validation

- `node test/run.js` ? 1,453 passed; the only failure is the pre-existing `package.json` 26.0.10 vs `CHANGELOG.md` 26.0.0 assertion
- Chrome and Firefox extractor modules are byte-identical
- `node --check` for both extractor modules ? passed
- `git diff --check` ? passed


esokullu:
Pushed a commit extending the recovery to one more case. No Copilot findings on this one — its review was a summary only.

**What I added.** Skipping unparseable candidates covered the illustrative-object case but not its neighbour: a `{` that never closed still abandoned the whole candidate, so a stray brace in prose ahead of the real object discarded it. Scanning now resumes at `start + 1` on an unbalanced opener, capped at 16 restarts to keep a pathological `"{{{{…"` response linear.

This is the same fix as #2682, deliberately. The two scanners had drifted apart on exactly this behaviour — each handled one half of the problem — which is worth knowing about since it's how the same bug ended up in both trees. They agree now.

**Worth noting for reviewers**, since the description mentions only the planner: this function also backs three classifier paths — `agent.js:11454` (progress-intent), `_isSafetyRefusalTerminal`, and `_looksLikePlanOnlyTerminal`. I walked all three and none looks like a downgrade; every consumer shape-checks (`normalizePlan` requires `summary`, the refusal detector requires `confidence === 0` plus refusal wording, the plan-only check requires ≥5 policy keys), and for the refusal detector the change makes a refusal *more* likely to be caught, which is the safe direction. Still, the blast radius is wider than "planner" suggests.

**Tests.** An unbalanced opener followed by two broken candidates then the real plan; the bounded brace flood; and one pinning the limitation this PR correctly scopes out — an earlier *valid* non-plan object still wins and returns `null`. That last one asserts current behaviour on purpose, so if anyone later moves to "first object that normalises," the test fails and forces the decision to be explicit rather than silent.

1453 passed, security corpus 60/60, Chrome and Firefox copies `cmp`-verified identical. The one failure is the pre-existing `package.json` 26.0.10 vs CHANGELOG 26.0.0 mismatch.

alectimison-maker:
Thanks. I reviewed the current head (`11c1431e`) and agree that recovery from an unterminated opener belongs with the invalid-candidate fix. Restarting at the next candidate, with the 16-restart cap, preserves recovery without leaving malformed brace floods unbounded.

I also traced the extractor's planner, classifier/terminal, and visible-media consumers. Their existing shape checks remain in place, so the scanner change makes later parseable JSON reachable but does not make an arbitrary object a valid plan, refusal, plan-only result, or media location. The known first-valid-object limitation remains unchanged.

I've updated the PR description to include the unbalanced-opener behavior and the wider consumer set. Local validation is 1,453 passing runner tests with only the existing package/CHANGELOG mismatch, 60/60 security-corpus checks, syntax checks, Chrome/Firefox parity, and `git diff --check`.

PR #2699, opened 2026-08-08
feat: add value-free teacher mode

### LINKED ISSUE #197: teacher mode
teacher mode so webbrain learns how to automate

alectimison-maker:
## Summary

- add `/teach --start <name>`, `/teach`, and `/teach --end`
- capture trusted user demonstrations in a tab-scoped temporary session
- compile demonstrations into the existing `webbrain-workflow/1` format
- reuse the saved workflow store and guarded replay executor
- add Chrome/Firefox parity, localized UI messages, and documentation

## Design

Teacher mode captures trusted top-document clicks, field completion,
checkbox/radio changes, Enter submissions, and navigation.

Field values are never read or stored. A demonstrated field is converted
directly into a runtime workflow parameter at the capture/compiler boundary.
URLs are reduced to sanitized origins, path families, and query-free replay
destinations.

Temporary sessions live in `storage.session`, survive navigation and service
worker restarts, and are always removed by `/teach --end`, including failed
compilations. Unsafe targets and actions beyond the workflow limit are skipped
with explicit warnings.

This introduces no new workflow schema or browser permissions.

## Scope

- captures the top-level document only
- cross-origin iframe demonstrations are intentionally deferred
- supports clicks, text/select fields, checkbox/radio state, Enter submission,
  and explicit navigation
- limits demonstrations to 100 recorded actions

## Testing

- `node test/run.js`: 1518 passed, 1 existing upstream failure
  - existing failure: package version `26.2.2` is newer than the latest
    changelog entry `26.2.0`
- security injection corpus: 60/60 passed
- rich-text toolbar guard: 33 passed
- Chrome/Firefox syntax and parity checks passed
- trusted-event browser harness passed for fields, checkboxes, Enter submission,
  pending-field flush, duplicate-submit suppression, and value non-retention
- contenteditable privacy harness confirmed entered text is not retained

Closes #197


webbrain-one:
Review finding: **request changes before merge**.

Teacher mode prevents `/teach --start` when an agent is already running, but it does not prevent an agent, scheduled job, cloud run, or saved-workflow run from starting after the teacher session is active. The capture script treats any `event.isTrusted` input as a user demonstration, while WebBrain deliberately dispatches CDP mouse/key events that are `isTrusted`. Those automated actions and navigations can therefore be recorded and compiled as if the user demonstrated them, which breaks the PR's trusted-user-demonstration boundary and can produce an incorrect saved workflow.

Suggested fix:

- Enforce teacher/run mutual exclusion at a central run-entry boundary covering interactive, scheduled, cloud, and workflow runs; either reject the run or explicitly stop the teacher session before automation begins.
- Also guard `record_teacher_action` and teacher navigation recording while an agent owns the tab, as defense in depth.
- Add a regression test for: start teacher mode → begin an automated run → dispatch trusted click/type/navigation → confirm no automated action is retained (or confirm the run is rejected).

Validation otherwise looks good: local `test/run.js` reached **1518/1519**; the sole failure is the existing `package.json` 26.2.2 vs `CHANGELOG.md` 26.2.0 mismatch on `main`. The security corpus passed **60/60**, CI passed **14 scenarios**, cloud-capture passed, and GitHub's WebMCP smoke check passed.

Since #2698 has now merged, this branch also needs a rebase/conflict resolution that preserves both the workflow-manager UI and the `/teach` additions.

webbrain-one:
Implemented the requested teacher/run exclusion and rebased the PR onto current main while retaining the teacher-mode work.

What changed:
- Added one agent run-entry guard used by interactive, streaming, scheduled, cloud, and saved-workflow runs. An active teacher session now rejects the run with teacher_mode_active before agent automation starts.
- Scheduled and cloud paths reserve the target tab and invoke the same guard before scheduled URL navigation, cloud tab activation, or capture setup.
- Teacher startup is serialized against those run reservations, closing the start/run race.
- record_teacher_action and teacher navigation now return agent_running without retaining the event whenever interactive, detached, scheduled, cloud, or agent execution owns the tab.
- Added a Chrome/Firefox regression that starts teacher mode, attempts interactive/scheduled/cloud/workflow/streaming runs, verifies zero automated dispatch, then injects click/type/navigation capture while agent-owned and verifies the session still has zero actions.

Validation:
- Node suite: 1,533 tests passed, including the new regression and scheduler/cloud tests. The only local failure is the packaging assertion for the absent untracked dist/webbrain-chrome-27.0.0.zip artifact.
- Security corpus: 60/60 passed.
- Rich-text toolbar guard: 33/33 passed.
- CI unit/cloud-capture scenarios passed.

PR head is now 38beb4ad and GitHub reports it mergeable; the WebMCP smoke check is queued.

webbrain-one:
Follow-up: the WebMCP smoke check completed successfully (22s). The only remaining failing status is Vercel's external authorization-required check.

PR #2747, opened 2026-08-11
feat(agent): resolve visual targets to semantic refs before coordinate fallback

feiniao87968492:
## Summary

- add a read-only `resolve_visual_target({ x, y, from_screenshot })` agent tool
- resolve screenshot-selected points to stable semantic accessibility refs across composed ancestors and open shadow DOM
- preserve canonical CSS coordinates for deliberate fallback through the existing coordinate click tool
- mirror Chrome and Firefox behavior, hide WebBrain indicators during hit testing, and classify page-authored results as untrusted

## Testing

- `node test/run.js` — 1622 passed, 0 failed
- `npm test` — passed before rebasing onto the latest upstream main
- `git diff --check` — passed
- Playwright fixtures added for nested SVG/button targets, canvas fallback, and open shadow DOM; local execution requires Playwright browser binaries


esokullu:
thank you. here's the inspection results:

<p>It genuinely helps when a visual point lands on an SVG/icon inside a button or inside open shadow DOM. Converting that point into a stable <code dir="ltr">ref_id</code> enables the safer <code dir="ltr">click_ax</code> path. The implementation is also thoughtful about coordinate conversion, indicator hiding, Chrome/Firefox parity, and untrusted page-authored labels.</p><p>But exposing it as another LLM tool creates avoidable overhead:</p>


Effect | Impact
-- | --
Tool schema and prompt guidance | A few hundred additional prompt tokens on eligible Act calls
Tool choice | Another option the model must understand and select correctly
Execution | An extra full model turn: inspect_viewport → resolve_visual_target → click_ax
Fallback cases | Canvas, iframes, labels/siblings, and slight coordinate misses often return the original CSS point, making that extra turn accomplish nothing



The extra inference turn is the real cost—not the schema tokens.</p><p>There are also limits hidden by the wording: it does not find the “nearest” target. It only checks the exact hit element and its composed ancestors. It won’t semantically recover many near misses, associated label/input siblings, iframe contents, or canvas controls. The existing coordinate-click path already performs screenshot scaling, form/select handling, label/input redirection, trusted CDP dispatch, and progress verification in [agent.js (line 16800)](/Users/barack/Documents/webbrain/src/chrome/src/agent/agent.js:16800).</p><p>My recommendation:</p><ul><li>Keep the semantic hit-testing implementation.</li><li>Do not expose <code dir="ltr">resolve_visual_target</code> to the model.</li><li>When the model calls existing <code dir="ltr">click({x,y,from_screenshot:true})</code>, internally:<ol start="1"><li>Convert coordinates once.</li><li>Hide indicators.</li><li>Attempt the semantic/shadow-DOM resolution.</li><li>Bind and click the resolved target through the existing safe path.</li><li>Fall back to the existing coordinate click when resolution fails.</li></ol></li><li>Return diagnostic metadata such as <code dir="ltr">semanticTargetResolved</code>, role, and bounded name.</li></ul><p>That gets almost all the benefit with no new tool choice and no additional LLM round trip. If internal integration proves too invasive, second-best is exposing the resolver only dynamically after a visual inspection—not in every Act tool catalog.</p><p>The fixtures are good mechanical tests, but I’d also want a small model-level evaluation comparing success rate, latency, tool calls, and tokens across nested SVG, shadow DOM, canvas, iframe, label/input, and dynamic rerender cases.</p>

feiniao87968492:
Thank you for your guidance!I'll make corrections in this direction

feiniao87968492:
I updated this based on your suggestion.

`resolve_visual_target` is no longer exposed to the model. Screenshot clicks now try semantic resolution internally, and fall back to the existing coordinate click when needed.

I also manually checked the main cases:

* SVG button → semantic path
* open shadow DOM button → semantic path
* canvas → coordinate fallback

Screenshots below show the page result and the Service Worker diagnostic.
canvas (2)
shadow (2)
svg (2)

esokullu:
thank you that was great! merged!

PR #2754, opened 2026-08-11
feat: localize selection shortcuts

### LINKED ISSUE #2746: Feedback from Chrome store: translate context menus
“As a Chinese user, for example, when I select a paragraph on X today and ask WebBrain to explain it, clicking “?” only shows options in English, such as “explain” and “Quiz me.” WebBrain has clearly recognized that I am a Chinese user, so could these options be displayed directly in Chinese? This causes the system to automatically respond in English after I click “explain,” and I have to ask it to answer in Chinese, after which it translates the response again. This wastes my tokens and is really not worthwhile.”

SO.. we need to translate context menus to the selected language

alectimison-maker:
## Summary

- localize the floating selection shortcut and native selection context menus for all 23 WebBrain interface locales
- rebuild native menus when `wbLocale` changes and apply live locale/RTL updates to the closed-shadow selection surface
- carry the interface locale into fixed selection actions so responses follow the selected UI language
- keep action IDs structural, use English per-key fallback, and preserve the existing `selection_only` grounding and nonce-based untrusted-content boundary

Closes #2746.

## Why

The extension already stores the selected interface locale, but the two selection-menu surfaces hardcoded English labels. Fixed actions also omitted that locale when building their model prompt, so a Chinese UI could still produce an English explanation. The implementation resolves one static, offline bundle in the background for both surfaces and adds the allowlisted response-language instruction outside the selected page-content boundary.

## Test plan

- `npm test` — passed (`1624 passed, 0 failed`; security corpus `60/60`)
- `npm run test:fixtures` — new Chrome/Firefox localization, live-switch, RTL, and fixed-action language cases pass; overall `149 passed, 2 failed`, with the same two pre-existing selection-dialog keyboard-focus failures as clean `main` (`147 passed, 2 failed` before these two new cases)
- `npm run test:webmcp` — both functional smoke assertions print `PASS`; the process then hits the existing 30-second cleanup timeout
- `npm run test:ci` — still hits the existing workflow/source assertion mismatch in `ci/test.mjs`; this PR does not modify either file
- `git diff --check`

> This PR was implemented and tested with AI assistance. The linked issue comment contains the AI triage disclosure and root-cause analysis.


webbrain-one:
Great addition! There are still a couple of open items, if we can close them this'll get merged right away! Thanks!

PR #2762, opened 2026-08-12
feat: add general knowledge selection scope

### LINKED ISSUE #2751: Allow selected-text custom questions to opt into general knowledge
> *This was generated by AI during triage.*

## Classification

Enhancement / design discussion. The current behavior is intentional and security-motivated, but it blocks a common selected-text question shape.

## User scenario

A user selects text that mentions or implies a concept, then types a trusted custom question such as:

> 这里的跨平台框架我很感兴趣，现在有哪些跨平台框架？

The selected excerpt discusses the idea of cross-platform frameworks but does not enumerate names. WebBrain responds that the selection is insufficient and asks the user to reselect text or start a new conversation, even though the model could answer the user-authored general question from its own knowledge.

## Current behavior and root cause

On current `main` (`9aa969c3`), `buildSelectionPrompt(..., 'custom', question)` applies the same source contract as fixed actions:

```text
Use only the text inside the selection block as source material for this action.
If the selection is insufficient, say so and ask the user to select more text.
```

The run is then structurally marked `selection_only`. `Agent.selectionGroundingScopes` persists that boundary across follow-up turns, disables browser tools, hides earlier conversation/page context, and appends a system note instructing the model not to answer beyond the selection. The only ordinary UI exit is New conversation.

This correctly protects fixed actions such as Summarize and Proofread from inventing source claims, but it conflates two different intents:

1. **Answer from the selection only**.
2. **Use the selection as an untrusted reference, while answering the trusted user question with general model knowledge**.

The distinction is especially visible because the popup already separates fixed action buttons from a custom “Ask WebBrain…” field.

## Design direction for discussion

Introduce an explicit structured grounding policy rather than weakening `selection_only`. For example, a `selection_context` policy could:

- keep the selected page bytes inside `<untrusted_page_content>` and never treat them as instructions;
- keep the user-authored question trusted;
- allow intrinsic model knowledge when answering that question;
- continue excluding the live page, screenshots, tools, attachments, and pre-selection conversation unless separately authorized;
- make the chosen scope visible in the UI and durable for follow-ups.

The safest UX is an explicit choice near the custom question field (for example, “Selection only” versus “Selection + general knowledge”), with `selection_only` as the conservative default. Another possible product choice is to make only the custom Ask field use `selection_context`, while all fixed action buttons remain `selection_only`; that is simpler but changes the default semantics and should be a maintainer decision.

## Acceptance criteria

- [ ] A user can explicitly ask a general-knowledge question anchored to selected text without clearing the conversation.
- [ ] Selected page text remains wrapped and treated as untrusted data.
- [ ] The broader mode does not silently expose the live page, browser tools, screenshots, attachments, or pre-selection history.
- [ ] Fixed source-bound actions retain their current selection-only guarantees.
- [ ] The active scope and the way to leave it are clear in the UI.
- [ ] Follow-up turns preserve the selected scope policy without accidental escalation.
- [ ] Chrome and Firefox behavior and regression coverage remain aligned.

## Out of scope

- Automatically inferring broader scope from page text.
- Weakening prompt-injection protections around selected content.
- Giving a selection-scoped run browser-action capabilities.

alectimison-maker:
## Summary

- add an explicit, default-off “Use general knowledge” option for custom selected-text questions
- carry a durable `selection_context` policy through prompt transport, retries, persistence, follow-ups, and the scope banner
- keep selected content untrusted and continue excluding the live page, screenshots, tools, attachments, and pre-selection history
- preserve `selection_only` for every fixed selection action and legacy conversation state
- ship matching Chrome/Firefox behavior and all 23 interface translations

## Design

The broader policy permits only the model intrinsic knowledge in addition to the selected text. It does not widen browser or conversation access. Unknown policies fail closed, and only an exact custom-action opt-in can select the broader scope.

The selection popup uses a two-column fixed-action grid so the custom question remains reachable in narrow viewports after adding the scope control.

## Testing

- `npm test` — 1641 main tests, 33 toolbar-guard tests, and 60 security checks passed
- `npm run test:fixtures` — 153 Chrome/Firefox browser fixtures passed
- `git diff --check`

## Compatibility and risk

Legacy persisted selection scopes safely hydrate as `selection_only`. The policy is structural rather than inferred from prompt text, and both browser builds use the same allowlist and UI behavior.

Closes #2751


fly1d (commented):
I found one defense-in-depth gap in the final policy boundary.

`Agent._selectionGroundedRunOptions()` accepts any allowlisted `sourceGrounding` independently of `selectionAction`. For example, at this PR's head:

```js
const options = agent._selectionGroundedRunOptions(
  1,
  [{ role: 'system', content: 's' }],
  { sourceGrounding: 'selection_context', selectionAction: 'summarize' },
);
```

produces and persists:

```json
{"sourceGrounding":"selection_context","selectionAction":"summarize"}
```

The popup/background path does currently emit `selection_context` only for custom Ask, so this is not reachable through the normal UI. However, the agent is the authoritative durable boundary (including retries/restores), and the PR description says only an exact custom-action opt-in may select the broader policy. A malformed or future transport path can therefore broaden a fixed action instead of failing closed.

Could `_selectionGroundedRunOptions()` downgrade `selection_context` to `selection_only` unless the normalized opening action is `custom`, and add a regression test for forged `selection_context + summarize` metadata?

alectimison-maker:
Addressed in `9f71b69a`.

The agent now normalizes the action first and downgrades `selection_context` to `selection_only` unless that normalized action is exactly `custom`. The same invariant is applied when hydrating or reusing durable scopes, so forged or legacy contradictory state is repaired before reuse.

Added a Chrome/Firefox regression covering both forged opening metadata and restored forged scope state.

Validation:
- 33/33 rich-text toolbar guard tests
- 1642/1642 main tests
- 60/60 security checks
- 153/153 fixture tests
- GitHub WebMCP smoke check passed

PR #2765, opened 2026-08-12
Fix #466: Expand sidepanel into standalone window

### LINKED ISSUE #466: Expand button at the top of the sidepanel
When clicked, Expand will open a *new* window, in Ask mode, but with the same provider configs as the sidepanel. This would be like the regular chatgpt user interface but especially for local model use. I know both Lm Studio and llama.cpp does provide this UI perfectly well, but what if it's vlllm, sglang, or ollama, and the user is looking for a quick easy way to chat with the LLM on a completely separate topic than what's on the web. We already have the plumbing, so a mini icon in the top of the sidepanel is too litle of a real estate and it will open this window for a convenient and almost full screen chatting experience.

Same btw with APIs. If someone is willing to test the chatting limits of an API, this would be a better way than the sidepanel.

Pls note, I'll be very picky with the UI and about the fact this won't break any of the existing functionality or bloat the existing sidepanel - which is still our core focus.

HeyShinde:
Resolves #466.

Adds a new "Expand" button to the sidepanel header next to the Settings and History buttons. 

When clicked, the button pops out the sidepanel into a standalone floating window, which makes it much easier to keep the chat open alongside your work without it being anchored to a specific browser tab. 

A few implementation details:
- Verified to work flawlessly on both Chrome and Firefox.
- The new window opens with the `?mode=ask&standalone=true` query parameters.
- If the sidepanel detects it's already running in a standalone window, the expand button hides itself so you don't accidentally spawn an infinite chain of popups.
- Added fallback localization strings across all 24 supported languages to keep the build green.

Tested locally, visually verified the UI matches the theme, and verified the test suite is green.


esokullu:
😍 thank you! will review this asap

PR #2821, opened 2026-08-16
test(agent): pin image-space coord-click bucketing under screenshot downscaling

alectimison-maker:
## Summary

Follow-up to #2813 (closed after review). Adds one test pinning the intended behavior: `_checkCoordClickLoop` buckets **image-space** coordinates — the space the model actually reasons in — because the screenshot click scale is always ≥ 1 (`cssW / shrunk.width`, set only when a resize happened), so converting to CSS pixels would magnify 2-4px image nudges out of the fixed 5px bucket and let a dead-button loop burn the whole step budget.

The test asserts both sides: image-space taps within a 5px bucket accumulate to a nudge on the 5th repeat, and the same taps converted at a realistic scale (2.449) spread across multiple CSS buckets.

## Testing

- `node test/run.js` — 1772 passed, 0 failed

### DIFF
--- a/test/run.js
+++ b/test/run.js
@@ -9428,11 +9428,35 @@ test('coord click: 5px drift collapses to same bucket', () => {
   assert.equal(d._checkCoordClickLoop(1, 102, 199).kind, 'nudge');
 });
 
-test('coord click: 10px drift = different bucket', () => {
+ test('coord click: 10px drift = different bucket', () => {
   const d = new ConfiguredLoopDetector();
   d._checkCoordClickLoop(1, 100, 200);
   // (115, 200) rounds to (115, 200) — different bucket
   assert.equal(d._checkCoordClickLoop(1, 115, 200).kind, 'none');
+});
+
+test('coord click: buckets in image space even when a downscale factor is stored', () => {
+  // The detector must measure the coordinates the model actually reasoned
+  // in — the image pixels of the screenshot it was shown. Screenshot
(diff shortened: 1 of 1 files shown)


test/run.js
```diff
@@ -9428,13 +9428,37 @@ test('coord click: 5px drift collapses to same bucket', () => {
...
 
-test('coord click: 10px drift = different bucket', () => {
+ test('coord click: 10px drift = different bucket', () => {
   const d = new ConfiguredLoopDetector();
   d._checkCoordClickLoop(1, 100, 200);
   // (115, 200) rounds to (115, 200) — different bucket
   assert.equal(d._checkCoordClickLoop(1, 115, 200).kind, 'none');
 });
 
+test('coord click: buckets in image space even when a downscale factor is stored', () => {
```
webbrain-one: **[P2] Exercise the scaled-click call path in this regression test**

The test feeds already-image-space values directly to LoopDetector and separately computes CSS buckets; it never stores a screenshot scale or invokes the Agent path that converts and dispatches click coordinates. A regression that forwards converted CSS coordinates to _checkCoordClickLoop would still pass. Exercise the actual click path with from_screenshot: true and a stored scale, or spy on the detector and assert that the original image-space arguments reach it.

webbrain-one (changes requested):
The added test does not yet protect the advertised coordinate-space invariant. Please exercise the Agent click path that owns screenshot scaling.

PR #2837, opened 2026-08-17
feat(ui): ask follow-up questions about selected answers

### LINKED ISSUE #2836: Design: ask follow-up questions about selected answer text
## Problem
When a WebBrain answer contains a detail that needs clarification, the user currently has to copy part of the answer manually and paste it into the composer. This breaks the reading-to-follow-up flow, especially for long streamed answers.

## Proposed design
- Allow users to select text inside an assistant answer in the side panel.
- Show a compact, accessible `Ask WebBrain` action anchored near the selection.
- Clicking the action should place the selected text into the existing composer as a clearly delimited quote, then focus the composer without sending automatically.
- Leave the quoted text editable so the user can add a question, correct the excerpt, or cancel it before sending.
- Dismiss the action when the selection is cleared, the user clicks elsewhere, or the conversation/tab changes.
- Keep the interaction local to the side-panel UI and mirror it in the Chrome and Firefox builds; no new background protocol or persistence format should be needed for the first slice.

## Acceptance criteria
- Selecting assistant answer text exposes the action without affecting normal text selection, copy buttons, or message metadata controls.
- The quote preserves the selected plain text and cannot inject markup into the side panel.
- The composer contains the quote and places the caret where a follow-up question can be entered.
- Clicking outside or clearing the selection removes the action without changing the composer draft.
- Keyboard and reduced-motion behavior remain usable.
- The behavior is covered by focused tests for quote construction and the selection-action lifecycle.

## Alternatives considered
1. Add another context-menu entry: discoverability is lower and it takes the user away from the answer they are reading.
2. Send the selected text immediately: it removes the opportunity to phrase a precise follow-up and is surprising for an interaction that starts as text selection.

The initial implementation should keep the quote representation simple and compatible with the current textarea composer; richer quote chips or cross-turn answer anchors can be follow-up work.

alectimison-maker:
## Summary

- Add an `Ask WebBrain a question` action when text is selected inside an assistant answer.
- Insert the selected text as an editable quote in the existing composer without sending automatically.
- Mirror the behavior across Chrome and Firefox.

## Motivation

Users currently need to copy part of a long WebBrain answer manually before asking a follow-up. This PR implements the design agreed in #2836 and keeps the reading-to-follow-up flow in the side panel.

## Design

The side panel observes selections and accepts only ranges contained in one assistant `.message-text` element. A localized, accessible floating button is positioned near the selection. Clicking it builds a plain-text Markdown quote in the existing textarea, preserves any existing draft after the quote, focuses the composer, and places the caret at the end. The action is coalesced onto one animation frame, rejects detached ranges, and is dismissed on selection changes, outside interaction, scrolling, resizing, send, Escape, tab changes, and conversation clears. No background protocol, persistence format, or model request changes are required.

## Testing

- `node test/run.js` — 1827 passed; 1 unrelated pre-existing failure: the stale Opera-safe flag license filename assertion expects a 32.1.0 archive while the checked-in archive is 32.0.0.
- `npm run test:toolbar-guard` — passed (33 tests).
- `npm run test:security` — passed (60/60 checks).
- `node --check` on both sidepanel and selection-quote modules — passed.
- Targeted full review against `upstream/main` — no blocking findings after fixing stale-range, rAF/layout, localization, and Escape lifecycle issues.
- Chrome MV3 unpacked extension check — passed: localized Chinese action rendered, selection action was clickable, quote appeared in the textarea, action dismissed, and console/errors were empty.
- Firefox manual browser check — unavailable in this environment.

## Compatibility and risks

This is a local UI-only change. Assistant text is inserted through the textarea value, not HTML, so selected markup remains literal text. The quote action does not auto-submit. Chrome was manually exercised; Firefox parity is covered by mirrored source and focused tests but was not manually exercised.

## Scope

Richer quote chips, persistent answer anchors, automatic send, and cross-turn answer navigation are intentionally deferred.

Closes #2836

### DIFF
--- a/src/chrome/src/ui/selection-quote.js
+++ b/src/chrome/src/ui/selection-quote.js
@@ -11,7 +11,8 @@ export function buildSelectionQuote(text) {
 export function buildSelectionComposerDraft(selectionText, draft = '') {
   const quote = buildSelectionQuote(selectionText);
   const existingDraft = String(draft == null ? '' : draft);
-  return quote ? `${quote}${existingDraft.trim() ? existingDraft : ''}` : existingDraft;
+  if (!quote || existingDraft.startsWith(quote)) return existingDraft;
+  return `${quote}${existingDraft}`;
 }
 
 export function selectionIsQuoteable({ startTextElement, endTextElement, text } = {}) {
--- a/src/chrome/src/ui/selection-quote.js
+++ b/src/chrome/src/ui/selection-quote.js
@@ -1,5 +1,48 @@
+const ELEMENT_NODE = 1;
+const TEXT_NODE = 3;
+const QUOTE_CHROME_TAGS = new Set(['BUTTON', 'INPUT', 'TEXTAREA', 'SELECT', 'SCRIPT', 'STYLE', 'TEMPLATE']);
+const QUOTE_CHROME_CLASSES = new Set(['code-block-header', 'code-copy-btn', 'code-lang', 'msg-copy-btn']);
+
 function normalizedSelectionText(text) {
   return String(text == null ? '' : text).replace(/\r\n?/g, '\n').trim();
+}
+
+function classListContains(node, className) {
+  if (node?.classList?.contains?.(className)) return true;
+  const classNameValue = typeof node?.className === 'string' ? node.className : '';
+  return classNameValue.split(/\s+/).includes(className);
+}
+
+export function isSelectionQuoteChrome(node) {
+  if (!node || node.nodeType !== ELEMENT_NODE) return false;
+  if (QUOTE_CHROME_TAGS.has(String(node.tagName || '').toUpperCase())) return true;
+  for (const className of QUOTE_CHROME_CLASSES) {
+    if (classListContains(node, className)) return true;
+  }
+  return false;
+}
+
+function collectSelectionQuoteText(node) {
+  if (!node) return '';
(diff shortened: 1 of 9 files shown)


esokullu:
great idea! looking into it.

esokullu:
## Review findings (fixed in bde18a66)

Pushed follow-up commit `bde18a66` (`fix(ui): harden selection quote follow-up action`) to address the review on this PR.

**P1 — Wait for pointerup before showing the quote action**
The chip was scheduled from `selectionchange` during an in-progress drag, and `positionSelectionAskAction` placed it 6px below the live range. A downward select could land on the button, extend the range out of `.message-text`, and dismiss the action. The chip now waits for `pointerup` / `pointercancel` (keyboard selection still uses `selectionchange` / `keyup`), and `.selection-ask-action` is `user-select: none`.

**P2 — Do not rewrite a live region on every refresh**
The button had `aria-live="polite"` and `textContent` was assigned on every refresh, including each animation frame while the selection changed. Removed `aria-live`. The label is written only when the locale or text actually changes.

**P3 — Exclude in-bubble chrome from quoted text**
`selection.toString()` included Copy buttons, code headers, and language labels inside `.message-text`. Quote text now walks the range and skips that chrome.

`node test/run.js`: 1828 passed. The remaining failure is the pre-existing Opera archive filename mismatch, unrelated to this change.

PR #2854, opened 2026-08-20
fix(chrome): release debugger sessions after runs

### LINKED ISSUE #2853: fix(chrome): release debugger sessions when a WebBrain run ends
## Problem

Opening or running WebBrain can leave Chrome showing that the extension is debugging the browser after the run has ended. The Chrome implementation stores debugger sessions by `tabId`, so CDP control is tab-scoped, but the session is not consistently detached when a run finishes, aborts, or a tab is closed.

This makes the scope unclear to users who want to use WebBrain in one tab while leaving other tabs unchanged. It also leaves the debugger attachment alive longer than the documented active-run lifecycle.

## Expected behavior

- Attach the debugger only to the target WebBrain tab that needs CDP.
- Keep other tabs free of WebBrain debugger attachments.
- Detach the target tab after the run or temporary CDP operation finishes, unless an explicitly active feature still owns the session.
- Clean up sessions when a run aborts, errors, or the tab is removed.
- Preserve existing per-tab Dev diagnostics, WebMCP, screenshot, and tool behavior.

## Scope

The first fix should centralize debugger-session ownership and cleanup by `tabId`, add regression coverage for normal and exceptional cleanup, and keep the broader question of reducing static `<all_urls>` content-script injection separate if needed.

alectimison-maker:
## Summary

- Add one tab-scoped CDP cleanup entry point that tears down diagnostics, WebMCP, and the debugger session in order.
- Release Chrome debugger sessions from both agent run paths and tab cleanup.
- Add regression coverage for cross-tab isolation and exceptional cleanup.

## Motivation

Chrome debugger sessions are stored by `tabId`, but the previous lifecycle did not consistently detach them after an agent run. That left the browser's debugging indicator active and made it unclear whether unrelated tabs were affected.

This addresses [#2853](https://github.com/webbrain-one/webbrain/issues/2853) without changing the broader static `<all_urls>` content-script design.

## Design

`CDPClient.cleanupTab(tabId)` is the single lifecycle seam. It attempts to disable Dev diagnostics and WebMCP, then detaches only the requested tab. Each teardown step is best-effort so a closing tab cannot prevent the final detach. The agent calls it from both non-streaming and streaming run `finally` blocks, and from tab cleanup. Existing CDP operations remain tab-keyed and are reattached lazily on the next operation.

## Testing

- `node test/run.js` — passed, 1895 tests
- `npm test` — passed, including 1895 unit tests, offline relevance benchmark, and 60/60 security checks
- `npm run test:ci` — passed, 14 scenarios and cloud capture checks
- `npm run test:webmcp` — passed with Chrome 151.0.7922.140
- Unpacked extension verification — passed in a dedicated headed Chromium session: extension enabled, settings page loaded from the real `chrome-extension://` origin, no console or page errors observed

## Compatibility and risks

- No manifest, storage, message, or tool schema changes.
- Firefox is unchanged because it has no Chrome CDP client or debugger permission.
- A subsequent run reattaches its target tab lazily as before.
- The broader reduction of static `<all_urls>` content-script injection remains separate work.

Closes #2853


esokullu:
Thanks for the debugger-lifecycle work. I pushed follow-up commit 41bb7518e to your branch to address the four review findings:

- WebMCP tool IDs now use a fresh opaque namespace for every registry generation, so stale conversation IDs cannot alias newly registered tools.
- Run cleanup preserves mode-scoped Dev console/network diagnostics across turns while full conversation/tab cleanup still drains them.
- Interactive, scheduled, and cloud runs retain the same-tab guard until asynchronous CDP cleanup completes.
- Successful and stopped deterministic saved-workflow replays now clean up their CDP state before releasing the run claim.

I also added focused regressions and updated the lifecycle documentation. The full npm test suite passes: 1,898 main tests, 33 toolbar-guard tests, and 60 security checks.

PR #2857, opened 2026-08-20
fix(settings): ignore stale model list responses

fly1d:
## Summary

- track model-load generations independently for each provider card
- stop superseded save, error, and model-list completions before they update the UI
- cover reversed response order against the real Chrome and Firefox settings functions

## Problem

A user can start loading models for endpoint A, change the provider settings to endpoint B, and load again. If B finishes first and A finishes later, the older A response currently overwrites the displayed model choices, repaired base URL, context window, and status for the newer B configuration.

## Testing

- `node test/run.js` (1895 passed, 0 failed)
- `npm run test:toolbar-guard` (33 passed)
- `npm run test:security` (60 checks passed)
- `node scripts/benchmark-offline-relevance.mjs`
- syntax and diff checks
- Chrome/Firefox model-load logic parity check

### DIFF
--- a/src/chrome/src/ui/settings.js
+++ b/src/chrome/src/ui/settings.js
@@ -3375,28 +3375,37 @@ function clearProviderLoadedModels(id) {
   if (datalistEl) datalistEl.innerHTML = '';
 }
 
+const providerModelLoadGenerations = new Map();
+
 async function loadProviderModels(id) {
   let datalistEl = document.getElementById(`models-${id}`);
   if (!datalistEl) return;
+  const generation = (providerModelLoadGenerations.get(id) || 0) + 1;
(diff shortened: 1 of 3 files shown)


esokullu:
Thanks for the original fix. I pushed a maintainer follow-up in 47d3f62e9 after finding one remaining overlap: the generation check ran only after saveProvider returned, so an older save could finish after a newer one and roll persisted provider settings back even though its stale model-list response was ignored. The follow-up serializes pre-load saves per provider, skips stale queued loads, and adds a Chrome/Firefox regression asserting that the newest settings persist last and only the current model request continues. Verification: 1,896 core tests, 33 toolbar guard tests, 60 security checks, and the offline relevance benchmark all pass; it also merges cleanly with current main.

PR #2889, opened 2026-08-22
fix(trace): repair runs interrupted by service-worker eviction

### LINKED ISSUE #2888: feat(trace): synthesize closing events for runs interrupted by service-worker eviction
## Motivation

Trace runs are persisted from a service worker that may be evicted between lifecycle events. The recorder can recover an event sequence number after restart, but it does not repair a durable run left with `status: "running"`. Such traces remain permanently in-progress in the Traces UI and lack a structured terminal cause.

## Scope

- On recorder initialization and when the Traces page opens, identify stale `running` trace records older than a conservative timeout.
- Append a synthetic structured `error` event describing service-worker interruption, continuing the existing run sequence.
- When turn/step lifecycle events are present, synthesize compatible closing lifecycle evidence as needed.
- Finalize the run record with an explicit repair marker (for example `repairedBy`) so repair is idempotent and inspectable.
- Add Chrome/Firefox mirror coverage for stale-run selection, sequence continuation, terminal event construction, and no duplicate repair.

## Constraints

- Keep default trace privacy unchanged: no prompt/message/tool content may be introduced.
- Reuse `trace/event-model.js`, the existing recorder queue, and `_peekSeq`; do not add a parallel trace format.
- Do not modify active/recent runs or rely on agent in-memory state.

## Acceptance criteria

- After a service-worker interruption and later restart, stale runs no longer remain permanently `running` in Traces.
- The repaired event log has a continuous sequence and an explicit structured interruption cause.
- Re-running repair does not append another terminal event or alter an already repaired run.
- Chrome and Firefox remain behaviorally mirrored.

alectimison-maker:
## Summary

- Repair stale trace runs left in `running` after a service-worker eviction.
- Append ordered, content-free structured interruption evidence and close any open step/turn lifecycle events.
- Apply the same behavior to Chrome and Firefox, from background startup and when the Traces page opens.

## Motivation

The recorder can persist a running run before the service worker is evicted, while the in-memory `endRun()` call is lost. The run then remains permanently `running`, even though its event sequence can still be recovered.

Closes #2888

## Design

- Scan durable runs with a conservative 10-minute stale threshold; recent runs remain untouched.
- Re-read each candidate and apply the repair plan in one `runs` + `events` IndexedDB transaction.
- Continue from the maximum durable event sequence, append `step_end`/`error`/`turn_end` evidence as applicable, and mark the run with `repairedBy` and `repairReason`.
- Reuse the existing event model and add the stable `SERVICE_WORKER_EVICTED` error code. No new event kind or trace schema migration is introduced.
- Keep Chrome and Firefox implementations mirrored.

## Testing

- `node test/run.js` — 1987 passed, 0 failed
- Node syntax checks for all touched JavaScript files — passed
- `node scripts/benchmark-offline-relevance.mjs` — completed successfully
- Manual Chrome unpacked-extension smoke test: extension loaded, Traces opened, Refresh clicked, no console/page errors

## Compatibility and risks

- Additive run metadata and error code only; existing completed traces are not modified.
- Default tracing behavior and privacy filtering are unchanged; repair events contain no user content.
- The 10-minute threshold intentionally protects recent runs, using the latest durable event timestamp as activity evidence. A run that stays completely quiet for longer than the threshold remains the intended repair candidate; the same service-worker instance is additionally protected by its in-memory active-run state.
- Firefox was covered by mirrored source, parity tests, syntax checks, and the shared unit suite; manual Firefox browser verification was not run.

### DIFF
--- a/src/chrome/src/background.js
+++ b/src/chrome/src/background.js
@@ -112,6 +112,7 @@ import {
 const providerManager = new ProviderManager();
 const apocalypseController = createApocalypseController(chrome);
 const VISION_OFFSCREEN_URL = chrome.runtime.getURL('src/offscreen/offscreen.html');
+void workflowTrace.repairStaleRuns().catch(() => {});
 
 function normalizeVisionDownloadState(state) {
   return {
--- a/src/chrome/src/background.js
+++ b/src/chrome/src/background.js
@@ -112,7 +112,11 @@ import {
 const providerManager = new ProviderManager();
 const apocalypseController = createApocalypseController(chrome);
 const VISION_OFFSCREEN_URL = chrome.runtime.getURL('src/offscreen/offscreen.html');
-void workflowTrace.repairStaleRuns().catch(() => {});
+// The stale-run repair scan waits a beat after wake so a run resuming from
+// eviction registers its in-memory state first; the Traces page can also
+// request an immediate scan via WB_TRACE_REPAIR_STALE_RUNS.
+const TRACE_REPAIR_STARTUP_DELAY_MS = 15_000;
+setTimeout(() => { void workflowTrace.repairStaleRuns().catch(() => {}); }, TRACE_REPAIR_STARTUP_DELAY_MS);
 
 function normalizeVisionDownloadState(state) {
   return {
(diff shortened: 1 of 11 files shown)


esokullu:
I reviewed the branch and pushed 5480bba6e with fixes for three issues found during review. Full suite passes (1987 tests).

**Unbounded scan on every wake (perf).** `repairStaleRuns` walked every run ever recorded (`listRuns({ limit: Number.MAX_SAFE_INTEGER })`) on each service-worker startup and on each Traces page load. A run's last activity is never older than its `startedAt`, so only records with `startedAt <= now - staleAfterMs` can qualify; the scan now walks just that slice of the `startedAt` index.

**Startup race against resuming runs.** The scan fired at module top-level, so when the worker woke specifically to resume an interrupted run, repair could race its first event write and mark it abandoned mid-flight. Two mitigations: the background defers its startup scan by 15s, and `_repairRunInTransaction` re-checks `_runState.has(runId)` right before writing, so a run that has re-registered itself in memory is never touched.

**Traces page could close out live-but-silent runs.** The page's recorder instance has an empty `_runState`, so a run alive in the background but quiet for more than the stale threshold (a hung LLM request, a long tool call) could be terminated with spurious repair events from the page while still running. The page now asks the background to own the scan via a new `WB_TRACE_REPAIR_STALE_RUNS` listener (both builds) and falls back to a local scan only when the background cannot be reached, since then nothing can be running anywhere. The fallback keeps the original behavior available for pages opened before the background first wakes.

Test assertions were extended to pin the bounded candidate scan, the deferred startup call, the message route, and the local fallback.

PR #2904, opened 2026-08-23
feat(providers): support optional local API keys

### LINKED ISSUE #2896: Support local providers with api-key
llama.cpp, LM-Studio and some others local providers can be used with api-keys but current UI does not allow to enter api-key for this endpoints.

k4its1t:
## Summary

- expose API-key configuration for llama.cpp, Ollama, LM Studio, and the other optional-auth local providers
- keep optional local authentication inside a collapsed Advanced section while leaving the required Local Proxy key visible
- send configured Bearer credentials from llama.cpp chat and streaming requests
- mirror the settings and provider changes across Chrome and Firefox

Closes #2896

## Verification

- `node --check` on all four changed runtime files
- `node test/run.js`: 2023/2025 pass, including the new Chrome/Firefox UI and request-header regressions; the two remaining failures are pre-existing release-artifact checks (`CHANGELOG.md` latest entry is 33.1.1 while `package.json` is 33.1.5, and `dist/webbrain-chrome-33.1.5.zip` is absent)
- `node scripts/benchmark-offline-relevance.mjs`
- `npm run test:security`: 60/60 checks pass
- `git diff --check`

### DIFF
--- a/src/chrome/src/providers/llamacpp.js
+++ b/src/chrome/src/providers/llamacpp.js
@@ -37,6 +37,12 @@ export class LlamaCppProvider extends BaseLLMProvider {
     return !!this.config.useCompactPrompt;
   }
 
+  _headers() {
+    const headers = { 'Content-Type': 'application/json' };
+    if (this.config.apiKey) headers.Authorization = `Bearer ${this.config.apiKey}`;
+    return headers;
+  }
+
   _buildRequestBody(messages, options = {}, stream = false) {
     const body = {
       messages: this._chatMessages(messages),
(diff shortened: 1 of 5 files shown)


webbrain-one:
Pushed f4edc1e fixing the false invalid-key warning from review.

**Problem:** adding `OPTIONAL_LOCAL_API_KEY_FIELD` to `ollama`/`lmstudio` activated `providerApiKeyWarning()` for them for the first time. Both providers ship seeded sentinel keys (`'ollama'`, `'lm-studio'`) that are shorter than `MIN_API_KEY_LENGTH = 12`, so every configured Ollama/LM Studio user got a persistent *"won't work without a valid key"* warning on render and on every save — even though those servers work fine with the sentinel Bearer values (which profile-sync already excludes as non-credentials).

**Fix:** both browsers' settings UI now normalize keys matching those known sentinels to empty before the length check (`DUMMY_API_KEYS` set next to `MIN_API_KEY_LENGTH`), so they're treated like an optional local key instead of an invalid one. Genuinely short user-entered keys still warn.

**Tests:** new assertions for both browsers in the settings test; full `node test/run.js` remains 2023 passed / 2 failed (the two pre-existing release-artifact checks).

PR #2910, opened 2026-08-24
feat(trace-export): add session-aware OTLP export compatibility

### LINKED ISSUE #2908: trace-export: define session-aware OTLP input and compatibility contract
## Problem

The OTLP converter currently accepts one `webbrain-trace/1` run export and derives a trace identifier from that run. This makes it impossible for a collector to reconstruct a persisted session containing several related runs, including delegated runs and cross-session parent relationships.

The Traces UI now exposes the persisted session and lineage fields needed for this association. The exporter needs a compatible input contract before its identifiers and parent relationships can be changed.

## Proposed input contract

Keep the existing single-run shape valid and add an optional session bundle shape:

```json
{
  "schema": "webbrain-trace/1",
  "session": { "sessionId": "session-id" },
  "runs": [
    { "run": { "runId": "run-id", "conversationId": "session-id" }, "events": [] }
  ],
  "exportedAt": 0,
  "exportedByWebBrainVersion": ""
}
```

The exact envelope name and whether the bundle is produced by the Traces UI or only consumed by the converter can be adjusted during review. The important compatibility requirement is that existing single-run files remain readable.

## Proposed OTLP semantics

- One logical session maps to one OTLP trace.
- Each run maps to one OTLP span with a deterministic span ID derived from its persisted run ID.
- A same-session `parentRunId` becomes `parentSpanId` only when the parent record is present and uniquely resolved in the bundle.
- A parent from another session is represented as an OTLP Span Link, with the persisted parent session and run identifiers retained as WebBrain attributes.
- Turn and step records remain span events for now; they do not become child spans.
- Missing parents, duplicate IDs, and cycles remain represented with lineage attributes or events rather than being silently discarded.
- Content remains excluded by default, while the existing explicit content option keeps its current behavior.

## Compatibility and tests

Please confirm whether the bundle should reuse `webbrain-trace/1` with optional fields or use a separate session-export schema. The implementation should preserve the current single-run fallback and add pure fixtures for:

- a root and same-session child run;
- a cross-session parent link;
- missing and duplicate parents;
- a cycle;
- a legacy single-run export without lineage fields.

The follow-up implementation should be limited to the pure converter and its tests. Provider `traceparent` propagation and step-level child spans remain separate work.

### LINKED ISSUE #2909: trace: define export version negotiation and legacy read compatibility
## Problem

WebBrain trace records already carry a run-level format version, and the exported JSON currently uses the `webbrain-trace/1` schema. New event kinds and optional run fields have been added over time, but the compatibility contract is not yet written down as one testable policy.

Without that contract, future readers and exporters could either reject useful historical runs or silently reinterpret fields from older data.

## Proposed compatibility contract

- Treat a run without `traceFormatVersion` as the legacy baseline version.
- Keep existing tolerant read behavior for unknown event kinds and missing optional fields.
- Define when a new event kind or field requires a format-version change.
- Define the difference between the IndexedDB record version, the run format version, and the exported JSON schema version.
- Preserve the existing single-run export shape while coordinating session bundles with the session-aware OTLP input contract.
- Keep privacy defaults unchanged; compatibility handling must not make content fields durable or export them by default.
- Do not perform destructive data migration as part of the first compatibility implementation.

## Test matrix

Add pure fixtures and reader/exporter assertions for:

- a historical run with no format version;
- a current run with all optional lineage and statistics fields absent;
- an event log containing an unknown event kind;
- newer optional fields that an older reader would not know;
- malformed optional values that should be ignored or normalized safely;
- legacy single-run JSON that remains readable and exportable;
- records that can be read without losing persisted lineage references.

## Questions for maintainer confirmation

1. Should the existing `webbrain-trace/1` schema remain the compatibility envelope with optional fields, or should a future exporter use a separately versioned schema?
2. Which changes require a `traceFormatVersion` increment: new event kinds, required fields, or only changes in field meaning?
3. Should unknown events remain visible as placeholders in every reader, or should exporters preserve them as generic events while the UI keeps the current placeholder behavior?

The implementation should stay focused on the pure normalization/compatibility seam and its tests. Storage migration, provider context propagation, and step-level child spans should remain separate follow-up work.

alectimison-maker:
## Summary

Extends the post-run trace converter with a compatible session-aware OTLP export path and documents the trace versioning contract.

Closes #2908
Closes #2909

## Changes

- Adds a pure normalization seam for the existing single-run export and the optional `{ session, runs }` bundle shape.
- Keeps legacy single-run OTLP output unchanged, including its root span and model/tool child spans.
- Maps a session to one OTLP trace and each persisted run to one deterministic span.
- Resolves unique same-session `parentRunId` values to parent spans, represents cross-session parents as Span Links, and preserves missing, duplicate, and cyclic lineage with attributes.
- Keeps turn and step activity as span events and preserves the default content-free export behavior.
- Preserves unknown event kinds as generic OTLP events instead of silently dropping them.
- Documents the separate storage version, run format version, and export envelope version policies.

## Compatibility

The `webbrain-trace/1` envelope remains valid for additive fields and event kinds. Missing or malformed run format versions use the legacy baseline, while semantic changes, newly required fields, or changed sequence/timestamp meanings require an explicit version change. No storage migration or provider context propagation is included.

## Validation

- `node test/run.js` — 2037 passed, 0 failed
- `node scripts/benchmark-offline-relevance.mjs` — completed successfully

### DIFF
--- a/docs/export-and-workflow-formats.md
+++ b/docs/export-and-workflow-formats.md
@@ -103,10 +103,17 @@ npm run trace:otlp -- webbrain-trace-example.json \
   --output webbrain-trace-example.otlp.json
 ```
 
+For a legacy single-run input, the output retains the existing root span with
+model-call and tool child spans. A session bundle uses one trace per persisted
+session, one span per run, same-session `parentRunId` parent links, and span
+events for turn/step activity. Cross-session parents are represented as span
+links rather than parent spans.
+
 The output is an
 [OTLP/HTTP JSON](https://opentelemetry.io/docs/specs/otlp/#json-protobuf-encoding)
-`ExportTraceServiceRequest`. It contains one `invoke_agent WebBrain` root span,
-child model-call and `execute_tool` spans, and lightweight lifecycle events.
+`ExportTraceServiceRequest`. Legacy input contains one `invoke_agent WebBrain`
+root span, child model-call and `execute_tool` spans, and lightweight lifecycle
+events; session bundles contain one `invoke_agent` span per run.
 The mappings follow the current
 [OpenTelemetry GenAI semantic conventions](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-agent-spans.md),
 which are still marked development and may change.
@@ -133,6 +140,10 @@ npm run trace:otlp -- webbrain-trace-example.json \
   --output webbrain-trace-example.otlp.json \
   --include-content
 ```
+
+See [`trace-format-compatibility.md`](trace-format-compatibility.md) for the
+relationship between the storage version, run format version, and export
+schema, plus the reader obligations for unknown events and legacy records.
 
 This is a post-run conversion, not live OpenTelemetry instrumentation. Child
 span start times are reconstructed from the recorder's completion timestamp and
--- a/docs/trace-format-compatibility.md
+++ b/docs/trace-format-compatibility.md
@@ -15,8 +15,10 @@ WebBrain trace data has three independent version layers:
 ## Reader obligations
 
 Treat a missing or malformed `traceFormatVersion` as the legacy baseline. Keep
-missing optional fields harmless. Readers must not fail an entire export because
-one event kind is unknown:
+missing optional fields harmless. Reject a numeric `traceFormatVersion` newer
+than the latest version the reader supports instead of interpreting `seq` and
+`ts` with older semantics. Readers must not fail an entire export because one
+event kind is unknown:
 
 - the Traces UI shows a labeled placeholder with the raw event view;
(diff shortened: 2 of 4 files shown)


esokullu:
Addressed both review findings in `25a1147f8`:

- Reject numeric `traceFormatVersion` values newer than the converter supports before sorting events or interpreting timestamps.
- Normalize each run's effective conversation ID before lineage grouping, so whitespace-only IDs fall back to the bundle session and same-session children use `parentSpanId` instead of a cross-session link.
- Added regression coverage for both cases and documented the future-version rejection rule.

Verification: `npm test` passes (toolbar guard, main suite, offline relevance benchmark, and security corpus).

PR #2912, opened 2026-08-24
feat(trace): export conversation trace bundles

alectimison-maker:
## Summary

- Extend the Traces page JSON export to include all persisted runs in the selected conversation.
- Keep standalone exports compatible with the existing `{ run, events }` shape and use `{ session, runs }` for conversation exports under `webbrain-trace/1`.
- Reuse the existing session index and screenshot loading path for every exported run.
- Apply lossless credential redaction to each sensitive run in a session bundle while preserving ordinary run content.
- Keep Chrome and Firefox behavior mirrored and document the export shapes, filenames, and privacy expectations.

## Compatibility

- No trace event or storage schema changes.
- Existing standalone JSON consumers retain their current envelope.
- The export remains local and does not upload data.

## Validation

- `node test/run.js` — 2041 passed, 0 failed

### DIFF
--- a/docs/export-and-workflow-formats.md
+++ b/docs/export-and-workflow-formats.md
@@ -87,7 +87,10 @@ changing the `webbrain-trace/1` schema:
 Lossless entries are redacted with the same credential-key safeguards as
 standalone JSON exports. Standalone exports keep the original `run` and
 `events` fields for existing consumers; session bundles use `session` and
-`runs` instead.
+`runs` instead. Before downloading a session bundle, the Traces page confirms
+the number of included runs and warns when any run uses the sensitive debug
+tier. A session export fails instead of silently producing a partial bundle if
+its run or event records cannot be read.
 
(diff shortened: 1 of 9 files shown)


esokullu:
Addressed the latest review findings in `a0b2d1380`:

- `trace-to-atif.mjs` now accepts session `{ session, runs }` bundles and converts them into one ordered multi-turn ATIF v1.7 trajectory with contiguous steps, source-run metadata, aggregated metrics, and collision-safe tool-call IDs.
- Session export now confirms conversation scope and run count, warns when any included sibling is a lossless/sensitive-debug run, and labels the export action as conversation-scoped.
- Session loading bypasses the default 500-run cap, and run/event read failures abort the export instead of producing a partial-looking JSON file.

Verification: `npm test` passes (33 toolbar guard tests, 2,042 main tests, offline relevance benchmark, and 60 security checks).

PR #2915, opened 2026-08-24
feat: add YouTube video loop suggestion

### LINKED ISSUE #2911: suggestion action on youtube: autoloop this video
it's a hidden feature of youtube;
webbrain already takes care of it but it's gotta be in dev mode
so the suggestion action button should take care of it too, switch to dev mode (aka enable js)

and perhaps write an adapter for weaker models to run this:

```
{
  "code": "return (()=>{ const v=document.querySelector('video'); if(!v) return {found:false};\n // ensure loop stays on even if YouTube resets it\n if(!window._ytLoopInterval){\n   window._ytLoopInterval = setInterval(()=>{ if(!v.loop) v.loop=true; }, 1000);\n   // also handle ended via YouTube player API if available\n   v.addEventListener('ended', ()=>{ v.currentTime=0; v.play().catch(()=>{}); });\n   // try YouTube player API loop\n   try{\n     const player = document.querySelector('#movie_player');\n     if(player && typeof player.setLoop === 'function') player.setLoop(true);\n   }catch(e){}\n }\n return {found:true, loop:v.loop, intervalSet: !!window._ytLoopInterval, paused:v.paused, duration:v.duration};\n})();"
}
```

Result

```{"success":true}```

w3lld1:
## Summary

- Adds a **Loop this video** suggested action on YouTube watch, Shorts, live, and short-link video pages in both Chrome and Firefox.
- Switches through the existing Dev-mode guard before sending a focused `execute_js` instruction.
- Keeps looping active when YouTube replaces the `<video>` element, while clearing the previous timer and event listener when the action is run again.

## Validation

- Focused Node behavior checks for Chrome/Firefox parity, idempotent reruns, video-element replacement, and ended-event replay
- `node --check` on both recommended-action modules and both side panels
- `git diff --check`
- `npm run test`: all 2,044 unit/regression tests passed, the offline-relevance benchmark completed, and all 60/60 security corpus checks passed

Fixes #2911

### DIFF
--- a/src/chrome/src/ui/recommended-actions.js
+++ b/src/chrome/src/ui/recommended-actions.js
@@ -473,40 +473,6 @@ export function buildRecommendedActions(pageInfo = {}, options = {}) {
         ['Call read_youtube_transcript for the active tab with timestamps:true, text_limit:6000, and include_segments:false.', 'Summarize the transcript with key points and timestamps when available.'],
       ),
     });
-    addUnique(actions, {
-      id: 'loop-youtube-video',
-      label: 'Loop this video',
-      prompt: `Use execute_js once with exactly this function body, then report whether a video was found and looping was enabled:
-const KEY = '__webbrainYouTubeLoop';
-const previous = window[KEY];
-if (previous?.timer) clearInterval(previous.timer);
-if (previous?.video && previous?.endedHandler) previous.video.removeEventListener('ended', previous.endedHandler);
-const state = { video: null, endedHandler: null, timer: null };
-const enforceLoop = () => {
-  const video = document.querySelector('video');
-  if (!video) return false;
-  if (state.video !== video) {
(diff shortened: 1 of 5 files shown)


esokullu:
Implemented the two review follow-ups in `65b6802cf`:

- Scoped YouTube loop enforcement to the original route, video element, and player. The interval and event listeners are removed, and loop state is disabled, when navigation or a player replacement changes the target.
- Prioritized the existing `download-media` recommendation ahead of `loop-youtube-video`, preserving the download shortcut when a YouTube comment composer is focused and the four-action limit applies.
- Mirrored the changes across Chrome and Firefox and added regressions for route changes, video replacement, cleanup, and focused-comment ordering.

Validation: `npm test` passed, including 33 toolbar-guard tests, 2,045 core tests, the offline retrieval benchmark, and 60 security checks.

esokullu:
thanks @w3lld1

Issue #2894, opened 2026-08-22
Add Pollination AI and Unsloth Studio integration


gaur-avvv:
## Summary
Add first-class integration support for Pollination AI and Unsloth Studio so WebBrain can use them as model providers and, where appropriate, offer browser-level guidance for their web interfaces.

## Motivation
- Pollination AI offers a broad OpenAI-compatible API for text, image, video, audio, embeddings, and realtime voice.
- Unsloth Studio exposes an OpenAI-compatible local/remote API for models loaded in its UI.
- Both can expand WebBrain’s provider coverage without requiring bespoke model adapters for every endpoint.

## Proposed implementation
1. Add a shared OpenAI-compatible provider abstraction for Pollinations and Unsloth.
2. Add provider settings for base URL, API key, model ID, timeout, and optional metadata.
3. Use secure storage for keys and avoid logging or exposing secrets.
4. Add a connection test using model discovery or a minimal request.
5. Add model discovery and capability metadata so the UI can distinguish chat, vision, embeddings, image, audio, and video support.
6. Add tests for streaming, tool calls, vision input, and common HTTP failures.

## Pollination AI details
- Base URL: `https://gen.pollinations.ai`
- OpenAI-compatible base: `https://gen.pollinations.ai/v1`
- Auth: `Authorization: Bearer <API_KEY>` or `?key=<API_KEY>`
- Model discovery: `GET /v1/models`
- Key prefixes: `sk_` for server-side, `pk_` for app/public flows
- New integrations should prefer Connect User Wallets / BYOP with OAuth authorization-code + PKCE or device flow
- Documented capabilities: streaming, tools, structured output, vision, search, embeddings, media inputs, image/audio/video/3D endpoints

## Unsloth Studio details
- Local OpenAI-compatible API at `http://localhost:<PORT>`
- Auth: API key created in Unsloth avatar → Settings → API
- Key prefix: `sk-unsloth-`
- Supported endpoints: `/v1/chat/completions`, `/v1/responses`, Anthropic-compatible `/v1/messages`
- Requirements: local Unsloth app running and a model loaded
- Remote access can be provided through Cloudflare HTTPS

## Acceptance criteria
- [ ] Provider settings UI for base URL, API key, model ID, and timeout
- [ ] Secure key storage with no secret leakage in logs, prompts, or screenshots
- [ ] Model discovery flow for both providers
- [ ] Chat completion support with streaming
- [ ] Tool calling and vision input where supported
[...]

PR #2923, opened 2026-08-26
feat(providers): add Unsloth Studio integration

khuzaymahbinharis-jpg:
## Summary

Addresses the **Unsloth Studio portion of #2894**, following the maintainer-approved split of Unsloth and Pollinations into separate work.

Adds Unsloth Studio as a first-class local OpenAI-compatible provider in WebBrain with Chrome and Firefox parity.

## Changes

- Add `unsloth` as a local OpenAI-compatible provider
- Support configurable Base URL, API key, and model selection
- Add authenticated `/v1/models` discovery
- Integrate with the existing Test Connection flow
- Reuse WebBrain's existing OpenAI-compatible chat, streaming, and tool-calling paths
- Keep vision support opt-in rather than assuming every loaded model supports vision
- Mirror provider and Settings behavior across Chrome and Firefox
- Add provider/security/parity regression coverage
- Document Unsloth Studio setup and credential/data-flow behavior

Pollinations support is intentionally not included in this PR.

## Validation

- `npm test` — **2,055 passed, 2 failed**
  - The same two failures reproduce on current upstream `main`:
    - changelog `33.2.1` does not match package version `33.2.2`
    - `dist/webbrain-chrome-33.2.2.zip` is absent
  - Current-main baseline: **2,053 passed, 2 failed**
  - All added Unsloth tests pass
- `npm run test:security` — **60/60 passed**
- `node --check` — passed on all touched JavaScript files
- `npm run build:zip` — Chrome, Edge, Firefox, and corresponding-source archives built successfully
- `git diff --check` — passed

## Notes

A live end-to-end smoke test requires an actual Unsloth Studio instance, generated API key, and loaded model, so that remains the only manual verification not performed locally.

Chrome and Firefox implementations are kept in parity.


webbrain-one:
Fixed both review findings in commit 661afab86:

- Unsloth model discovery now excludes catalog entries explicitly marked loaded: false, so Load Models only offers resident models under the default no-switching configuration. The behavior and regression coverage are mirrored in Chrome and Firefox.
- The Unsloth default Base URL is now http://127.0.0.1:8888/v1 in both provider managers, both Settings placeholders, and the provider documentation.

Verification:
- Syntax checks pass for both provider managers, both Settings files, and test/run.js.
- Both Unsloth regression tests pass for Chrome and Firefox, including resident-only discovery and the IPv4 default.
- The full test runner reports 2055 passed and 2 unrelated release-artifact failures: the newest changelog entry is 33.2.1 while package.json is 33.2.2, and dist/webbrain-chrome-33.2.2.zip is missing.

esokullu:
thank you and welcome to the club :D

PR #2948, opened 2026-08-29
feat: Make provider context windows and output token limits configurable

xianhongtao:
## Summary

- Add optional `Context Window` and `Max Output Tokens` settings for configurable LLM providers.
- Use the configured output limit throughout main generation, streaming, retries, and recovery.
- Add provider-specific defaults:
  - DeepSeek: 1,000,000 context / 384,000 max output
  - OpenAI: 272,000 context / 128,000 max output
  - Anthropic: 1,000,000 context / 128,000 max output
- Keep the legacy 4,096-token fallback for providers without an explicit output limit.
- Ensure duplicated providers do not unintentionally inherit output-limit overrides.
- Keep planner calls conservatively capped independently of the main generation budget.

## Rationale

The previous hard-coded 4,096-token output budget prevented providers such as DeepSeek, OpenAI, and Anthropic from using their supported output capacity.

OpenAI models support a larger context window, but prompts above 272K input tokens use long-context pricing. The OpenAI default is therefore set to 272K as a cost-aware baseline, while users can explicitly increase it in Settings.

## Testing

- Added regression coverage for provider context and output limits.
- Verified Chrome and Firefox configuration parity.
- Full test suite:
  - 2,110 passed, 0 failed
  - Security checks: 60/60 passed


esokullu:
Follow-up on this branch: `b24e68eba` — clamp card-wide max output tokens to the selected model's ceiling.

The shipped provider defaults (128k for Anthropic/OpenAI, 384k for DeepSeek) were merged onto every card and then forwarded on every main turn. That is correct for the default models, but a saved/selected model with a lower output limit (e.g. Claude Haiku 4.5 at 64k, GPT-4o at 16k, legacy `deepseek-chat` at 8k) still received the full card budget and could fail with a hard API validation error.

`resolveMaxOutputTokens()` now takes `min(configured budget, known model ceiling)` in both the provider getter and the agent's request path. Default models keep their full shipped budgets; a smaller Settings value still wins. Unknown models are not given an invented cap.

PR #2949, opened 2026-08-29
fix: restore Chinese Ask streaming copy

xianhongtao:
图片


esokullu:
thank you @xianhongtao I don't know when this happened. Very valuable contribution. Many thanks! Merging now.

PR #2955, opened 2026-08-30
feat(ui): add independent sidepanel scale controls

### LINKED ISSUE #2953: feat(ui): add independent zoom controls for the extension side panel
## What to build

Add an independent UI scale setting for the WebBrain side panel.

Browser page zoom currently affects the active web page but not the extension side panel. This makes the panel difficult to read on high-density displays and prevents users from sizing the page and extension UI independently.

The new setting should scale the complete side-panel interface, including text, controls, icons, spacing, dialogs, and the composer. It should apply to both the regular side panel/sidebar and its detached standalone window, while leaving settings, history, traces, readers, and other extension pages at their normal size.

Support both Chrome and Firefox.

### Scale levels

Use these fixed levels:

- 75%
- 80%
- 90%
- 100%
- 110%
- 125%
- 150%
- 175%

The default is 100%.

### Visible controls

Provide decrease, current percentage, increase, and reset controls in:

- The side-panel menu
- Settings → General → Appearance

Both locations update the same profile-wide setting immediately.

### Browser and extension zoom shortcuts

Keep browser page zoom and extension UI zoom independent:

- When the web page has keyboard focus, `Ctrl/Cmd + +/-/0` retains the browser's existing page-zoom behavior.
- When the extension panel has keyboard focus, the same shortcuts adjust or reset the extension UI scale.
- Merely hovering over the panel must not steal keyboard focus or introduce pointer-based shortcut forwarding.
- Preserve the browser's existing behavior when the pointer is over the panel but the panel has not been focused.

Because interception of native browser zoom accelerators is not guaranteed consistently across browser versions, visible controls must remain the reliable fallback.

Also expose dedicated global extension commands:

- `Alt+Shift+Comma`: decrease extension scale
- `Alt+Shift+Period`: increase extension scale
- `Alt+Shift+0`: reset extension scale

The settings page should show the current dedicated bindings and direct users to the browser's native extension-shortcut configuration UI when they want to change them.

## Persistence

Store the selected scale per browser profile.

The setting should:

- Apply immediately without reloading the panel
- Stay synchronized between the side panel and detached window
- Persist across browser restarts
- Be included in configuration export and import
- Fall back safely to 100% when the stored value is missing or invalid
- Apply before the panel becomes visible where possible, avoiding a 100%-to-saved-scale flash

## Acceptance criteria

- [ ] Chrome and Firefox expose matching UI-scale behavior.
- [ ] The complete side-panel UI scales, rather than only its text.
- [ ] The side panel and detached window share one scale value.
- [ ] Other extension pages remain at their normal size.
- [ ] Users can select every defined scale level from the panel menu and settings.
- [ ] Reset returns the scale to 100%.
- [ ] Controls stop or become disabled at the minimum and maximum levels.
- [ ] Scale changes are applied immediately and synchronized across open panel instances.
- [ ] The setting persists across browser restarts.
- [ ] Configuration export and import preserve the scale.
- [ ] Invalid imported or stored values safely resolve to 100%.
- [ ] `Ctrl/Cmd + +/-/0` adjusts the extension only when the panel has keyboard focus.
- [ ] The implementation does not steal focus when the pointer enters the panel.
- [ ] Dedicated extension commands work while focus is outside the panel.
- [ ] The settings UI exposes the current dedicated bindings and a route to the browser's shortcut manager.
- [ ] The composer, message list, menus, dialogs, tooltips, toasts, and scrolling remain usable at every scale level.
- [ ] Scaling does not introduce horizontal clipping or hide controls at supported side-panel widths.
- [ ] New controls have accessible labels and follow the existing localization system.
- [ ] Automated tests cover scale validation, persistence, configuration transfer, shortcut handling, and Chrome/Firefox parity.
- [ ] Native `Ctrl/Cmd + +/-/0` behavior is manually verified with a real keyboard in supported Chrome and Firefox versions.

## Non-goals

- Scaling the active web page through the extension
- Per-site, per-tab, or per-window scale preferences
- Scaling settings, history, traces, PDF/Wikipedia readers, or other extension pages
- Refactoring all fixed pixel sizes to `rem`
- Moving keyboard focus based on pointer hover

## Blocked by

None — can start immediately.

alectimison-maker:
## Summary

- Add profile-wide UI scale controls for the side panel and detached chat window in Chrome and Firefox.
- Keep extension UI scale independent from page zoom, with fixed levels from 75% through 175%.
- Add settings controls, configuration transfer support, focused keyboard handling, dedicated extension commands, and translated copy for every shipped non-English locale.

## Motivation

The browser page zoom does not resize the extension side panel, which makes the panel difficult to read on high-density or remotely viewed displays. This implements the behavior requested in [#2953](https://github.com/webbrain-one/webbrain/issues/2953) without changing page zoom or other extension pages.

## Design

- A mirrored pure helper normalizes the fixed scale levels, steps controls, handles focused `Ctrl/Cmd +/-/0`, and persists the canonical `uiScale` setting.
- The side panel and detached `sidepanel.html?standalone=true` share `chrome.storage.local`/`browser.storage.local`, with a synchronous localStorage pre-paint mirror and a canonical-storage reveal gate.
- CSS zoom scales the complete panel subtree; compensated height and viewport-coordinate conversion keep the composer and transient selection action usable at high scales.
- Chrome and Firefox expose the requested `Alt+Shift+Comma`, `Alt+Shift+Period`, and `Alt+Shift+0` commands, with settings links to each browser native shortcut manager.
- Chrome and Firefox locale sets use the correct `sp.*` and `st.*` namespaces; all new scale strings are translated in the 21 non-English locales, with English retained as the canonical source.

## Testing

- `npm test` — passed
- `node test/run.js` — 2123 passed
- Locale completeness and non-English translation regression assertions — passed
- `node --check` on changed JavaScript — passed
- Chrome unpacked extension at the real `chrome-extension://` origin — passed: settings and side-panel controls, 175% viewport/layout, simulated selection action positioning, Escape focus return, focused `Ctrl+0`, and no console/page errors.

Firefox real-browser verification and native keyboard verification in Firefox were not available in this environment.

## Compatibility and risks

- Existing profiles default to 100%; invalid stored or imported values fall back to 100%.
- Existing page zoom and non-side-panel extension pages are unchanged.
- The feature adds three browser command entries and preserves Chrome's four suggested-shortcut limit.
- The English locale remains the canonical source; all 21 non-English locale files now provide translated scale copy.

## Scope

- Browser-level manual verification for Firefox remains deferred.
- Full cross-browser visual coverage at every supported width remains for maintainer CI/manual review.

Closes #2953


alectimison-maker:
## Firefox verification

Completed against the PR branch with Firefox 153.0.4 using a temporary `web-ext` installation and the real `moz-extension://` origin.

Passed:

- Extension installed without Firefox add-on warnings.
- Side panel canonical-scale hydration and initial 100% rendering.
- UI-scale popover open/close behavior.
- Increase to 110%, including CSS `zoom` layout and the `localStorage` mirror.
- Focused `Ctrl+0` handling, including `preventDefault()` and reset to 100%.
- Settings-page increase control and shared storage persistence.
- `node test/run.js`: 2123 passed, 0 failed.
- Firefox-engine cases in `npm run test:fixtures`: all passed.
- Firefox JavaScript syntax checks: passed.

The aggregate fixture command still reports 175 passed and 1 failed because of an unrelated existing Chrome CDP rich-text keyboard audit (`ArrowDown` no-progress assertion); no Firefox case failed. Native OS-level shortcut dispatch and the full cross-width visual matrix were not exercised in this run.

webbrain-one:
Applied fixes from code review (commit \24e43b930\):

1. **Render only after persist** (\sidepanel.js\, \settings.js\): the UI scale render now happens after \saveUiScale\ resolves, so a failed storage write can no longer leave the displayed scale showing a value that wasn't persisted.

2. **Compensate width under CSS zoom** (\ui-scale.js\, \	heme-bootstrap.js\): \uiScaleLayout\ now inverse-compensates \width\ with the same \inverse%\ pattern it already used for \height\ (previously hardcoded \100%\). This prevents horizontal clipping of content past the panel edge at high scales. The pre-paint bootstrap was updated to stay in lockstep.

3. **Ignore shifted shortcuts** (\ui-scale.js\): added an \event.shiftKey\ guard so \Ctrl+Shift+=/-/0\ are not captured (native page zoom uses the unshifted combos).

Tests in \	est/run.js\ were updated and strengthened to cover the new width values, the shift-key guard, pre-paint width compensation, and render-after-persist ordering. Full suite: 2123 passed, 0 failed.

esokullu:
thanks @alectimison-maker all tech reviews are complete, just my manual review for visual inspection remains, I should do it soon.

esokullu:
Pushed a1e5c9013 to this branch via maintainer edit, with fixes from a review pass over the diff. Happy to split any of it into a separate PR if you'd rather keep this one focused.

**The width compensation in 24e43b9 was backwards.** This is the one worth a second opinion, since it undoes part of that commit. `zoom` shrinks the element's coordinate space, and a percentage width resolves against the containing block *inside* that space, so `width: 100%` already fills the panel and needed no compensation. Dividing it by the zoom left a 20% dead gutter at 125%. At 75% it went the other way and overflowed to 533px in a 400px panel, where `body { overflow: hidden }` propagates to the viewport and clips the settings and expand buttons with no scrollbar to reach them. Measured in Chromium and Firefox at 400×600, the body is now exactly 400×600 at all eight levels. The `vh` height compensation is correct and stays, since viewport units really aren't zoom-adjusted.

**Zoomed rects were mixed with unzoomed scroll coordinates.** `getBoundingClientRect()` returns zoomed viewport px, while `scrollTop`, `clientHeight` and `offsetTop` stay in the body's own CSS px. Four sites mixed the two, so they now convert through a shared `uiScaleZoom()` helper that `positionSelectionAskAction` reuses:

- `scrollChatToQuestion` overshot the question by up to 232px at 175% (measured before and after)
- `chatTurnNeedsReadingNavigation` compared an inflated turn span against `clientHeight`, so the reading chip appeared for turns that already fit
- `scrollSlashCommandOptionIntoView` scrolled past the option it was trying to reveal

**Scale hydration raced the write queue.** `loadUiScale` ran beside the queue rather than inside it, and the pre-paint value comes from the localStorage mirror that an MV3 worker can't refresh. So after a global shortcut with no panel open: storage holds 175, mirror holds 100, and a `Ctrl+=` before hydration lands writes 110 and silently drops the 175. The load now seeds the queue, so an early step waits for the canonical value instead of being dropped the way `settings.js` drops it.

**Two smaller ones.** The pin coachmark points at the toolbar pin icon, which sits outside the panel and doesn't move with the scale, so its offsets and size divide the zoom back out; at 175% the arrow was landing well below its target. And the scale popover chips had no hover feedback, because the two-id base rule outranks the generic `.header-right button:hover`.

Suite is green at 2123 passed. I updated the assertions that encoded the old width behaviour and added coverage for the zoom conversions, the queue seeding and the hover rule.

PR #2968, opened 2026-09-03
Fix broader context restoration after selected-text turns

### LINKED ISSUE #2961: clicked use broader context but it seems it didn't take effect
pls see traces. After the response, I click the use broader context button inside the sidepanel.[](url)

[webbrain-traces-1788237281115.md](https://github.com/user-attachments/files/31674116/webbrain-traces-1788237281115.md)

16Miku:
## Summary

- persist a one-shot restoration state after the user explicitly chooses broader context
- mark the next accepted turn with trusted runtime context and attach a fresh `read_page` result before the model answers, falling back to `get_accessibility_tree` when needed
- clear the restoration state after one use and when a new selection, conversation, independent run, or tab lifecycle supersedes it
- keep the Chrome and Firefox implementations in parity and add regression coverage

## Why

Clearing the selected-text boundary restored tool availability, but the next model turn could still follow the earlier selection-only conversation context without reading the page. This made **Use broader context** appear ineffective even though `selection_grounded` was already false.

The restoration is now explicit, persisted across service-worker restarts, consumed once, and backed by a deterministic fresh page read.

## Validation

- `npm test`
  - 2193 passed, 0 failed
  - selection scope restoration test passed for Chrome and Firefox
  - security corpus: 60/60 checks passed
- `node --check src/chrome/src/agent/agent.js`
- `node --check src/firefox/src/agent/agent.js`
- manual Chrome validation with WebBrain v34.1.4: after restoring broader context, the trace recorded `selection_grounded:false` and `read_page() -> success`; the answer used information outside the original selection

Closes #2961

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "test": "npm run test:provider-limits && npm run test:toolbar-guard && node test/run.js && node scripts/benchmark-offline-relevance.mjs && npm run test:security",
+    "test": "npm run test:provider-limits && npm run test:toolbar-guard && node test/run.js && node test/selection-scope-restoration.mjs && node scripts/benchmark-offline-relevance.mjs && npm run test:security",
     "test:provider-limits": "node test/provider-model-limits.mjs",
     "test:security": "node test/security/injection-corpus.mjs",
     "test:toolbar-guard": "node test/rich-text-toolbar-guard.mjs",
--- a/src/chrome/src/agent/agent.js
+++ b/src/chrome/src/agent/agent.js
@@ -14945,7 +14945,6 @@ Rules: no prose intro, no conclusion, no "this screenshot shows...", no layout d
       runOptions._standalonePersistedUserMessage = submittedUserMessage;
     }
     messages.push(submittedUserMessage);
-    this._consumeSelectionGroundingRestoration(tabId, submittedUserMessage);
     if (runOptions?.selectionGroundingScopeStarted === true) {
       this._finalizeSelectionGroundingScope(tabId, messages, submittedUserMessage);
     }
@@ -33285,7 +33284,6 @@ Rules: no prose intro, no conclusion, no "this screenshot shows...", no layout d
         return (finalResponse = error);
       }
       messages.push(enriched);
-      this._consumeSelectionGroundingRestoration(tabId, enriched);
       if (runOptions?.selectionGroundingScopeStarted === true) {
         this._finalizeSelectionGroundingScope(tabId, messages, enriched);
       }
@@ -33398,6 +33396,7 @@ Rules: no prose intro, no conclusion, no "this screenshot shows...", no layout d
       _traceStatus = responseOnly.status;
       return finalResponse;
     }
+    if (this._consumeSelectionGroundingRestoration(tabId, enriched)) this._persist(tabId);
     this._startPlanExecutionGuard(tabId, mode, gateOutcome, runOptions);
 
     if (this._isActionMode(mode) && !selectionOnly && !standaloneChatRun) {
@@ -34540,7 +34539,6 @@ Rules: no prose intro, no conclusion, no "this screenshot shows...", no layout d
     if (mode === 'dev' && provider.promptTier === 'compact') {
       const msg = this._devModeBlockedMessage(provider);
       messages.push(enriched);
-      this._consumeSelectionGroundingRestoration(tabId, enriched);
       if (runOptions?.selectionGroundingScopeStarted === true) {
         this._finalizeSelectionGroundingScope(tabId, messages, enriched);
       }
@@ -34596,6 +34594,7 @@ Rules: no prose intro, no conclusion, no "this screenshot shows...", no layout d
       );
       return finish(responseOnly.content, responseOnly.status);
     }
+    if (this._consumeSelectionGroundingRestoration(tabId, enriched)) this._persist(tabId);
     this._startPlanExecutionGuard(tabId, mode, gateOutcome, runOptions);
 
     let standaloneGroundingGap = this._standaloneOfflineGroundingGap(localWikipediaRag, runOptions);
--- a/src/chrome/src/agent/agent.js
+++ b/src/chrome/src/agent/agent.js
@@ -304,6 +304,7 @@ const SELECTION_CONTEXT_SCOPE_SYSTEM_NOTE = 'This conversation is anchored to te
 const SELECTION_CONTEXT_DIALOGUE_MESSAGE_CHARS = 6000;
 const SELECTION_CONTEXT_DIALOGUE_TOTAL_CHARS = 12000;
 const SELECTION_CONTEXT_DIALOGUE_MAX_MESSAGES = 12;
+const SELECTION_SCOPE_RESTORED_RUNTIME_NOTE = '[Selection scope status — TRUSTED WebBrain runtime state: The user explicitly removed the selected-text boundary. Any selection-only instruction in earlier conversation history is historical context, not a constraint on this user message. Normal access to the current page, browser tools, files, attachments, and the complete conversation is restored, subject to the usual mode and safety rules. This is the first accepted follow-up after that explicit restore, so WebBrain will attach a fresh read of the current page before the model answers whenever a page-reading tool is available. Interpret the latest request using the restored page and conversation context rather than treating the historical selected-text boundary as active.]';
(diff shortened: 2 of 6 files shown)


16Miku:
CI note: the `WebMCP E2E / smoke` failure timed out waiting for `fixture page error; received: []`. The same failure occurred on run 33682472957 for PR #2964, while subsequent `main` WebMCP runs succeeded, so this appears to be the existing flaky E2E rather than a failure in the selection-scope changes. I attempted to rerun the failed job, but GitHub requires repository admin permission.

The final rebased commit passes local `npm test`: 2193/2193 tests, the new Chrome/Firefox selection-restoration regression test, and 60/60 security checks.

The Vercel status points to the repository owner's Vercel authorization flow and is unrelated to this extension runtime change.

esokullu:
Read through this against `main`, Chrome and Firefox side by side. The shape of the fix looks right to me: tab-scoped one-shot state, persisted through `_persist`/`_hydrate`, cleared on the same lifecycle events as `selectionGroundingScopes`, with the deterministic read rather than hoping the model decides to call `read_page` on its own. Four things came up in review. Two of them will break real setups.

### 1. `webbrainSelectionScopeRestored` never gets stripped before the provider call

`src/chrome/src/providers/base.js:150`, `src/firefox/src/providers/base.js:149`

`_mapMessages` exists to keep our internal `webbrain*` markers out of the request body, and the new flag was not added to it. So it goes out as an unknown property inside the `messages[]` entry of the Chat Completions / Responses payload. Azure OpenAI, vLLM and other pydantic-validated gateways reject unknown message properties with a 400, which means the first turn after **Use broader context** fails outright on those endpoints. This is the same failure `webbrainPlannerClarification` was patched for in 627b65f72.

```diff
         !Object.hasOwn(message, 'webbrainPlannerClarification')
         && !Object.hasOwn(message, 'webbrainAppOwned')
         && !Object.hasOwn(message, 'webbrainAppOwnedKind')
+        && !Object.hasOwn(message, 'webbrainSelectionScopeRestored')
       )) {
         return message;
       }
@@
         webbrainPlannerClarification: _plannerClarification,
         webbrainAppOwned: _appOwned,
         webbrainAppOwnedKind: _appOwnedKind,
+        webbrainSelectionScopeRestored: _selectionScopeRestored,
         ...providerMessage
       } = message;
```

### 2. The new note breaks the anchored match in `_progressPageScopeFromConversation`

`src/chrome/src/agent/agent.js:21282`, firefox `:18884`

That loop matches `/^\s*\[Current page context[^\]]*\bURL:/` after `stripTrustedRuntimeContext`, which only removes the runtime-context block. `SELECTION_SCOPE_RESTORED_RUNTIME_NOTE` lands between the runtime context and the page-context line, so on a restored turn the anchor fails and the loop keeps walking back to an older message.

Concrete case: selection-scoped turn on page A (selection-scoped turns skip URL/title, so no page context is recorded), user restores broader context on page B, then starts an agent task. The progress ledger scopes its task keys to a stale URL, or to nothing.

```diff
-      const c = stripTrustedRuntimeContext(this._messageText(messages[i]?.content));
+      const c = stripTrustedRuntimeContext(this._messageText(messages[i]?.content))
+        .replace(/^\s*\[Selection scope status[^\]]*]\s*/i, '');
```

You already added the same strip to `_stripPinnedNotePrefixes` at `:20815`, so this is the one call site that got missed.

### 3. Fixed synthetic tool-call id

`src/chrome/src/agent/agent.js:14871`, firefox `:12663`

`id: 'selection_scope_restored_first_read'` is a constant, but a conversation can restore broader context more than once (select text, ask, restore, select new text, ask, restore). Both synthetic calls then sit in the same `messages` array and go out in the same request with identical ids on two separate `tool_calls`/`tool_result` pairs. Suffixing with `messages.length` at injection time is enough:

```diff
-      id: 'selection_scope_restored_first_read',
+      id: `selection_scope_restored_first_read_${messages.length}`,
```

### 4. Restoration is consumed before the turn is known to proceed (worth a maintainer opinion)

`src/chrome/src/agent/agent.js:14945`, firefox `:12731`

`_consumeSelectionGroundingRestoration` runs inside `_maybeRunPlannerGate`, right after the user message is pushed. If the read-scope classifier then returns `proceed:false` (`:14958`), or the planner comes back `responseOnly:true` (`:33386`), the one-shot flag is already gone and the deterministic read never happens. The user answers the clarifying question and that follow-up turn behaves as if broader context was never restored, which is the bug this PR is fixing.

I left this alone because the PR describes consumption as happening "after message acceptance" and the regression test asserts it, so it reads as a deliberate choice. The correction note does reach the model on the gated turn. Still, moving the consume to the point where the turn actually commits to running would close the gap, and it seems worth a decision rather than leaving it implicit.

### One unrelated nit

The raw-image hunk in `_buildInitialUserMessage` left the `enrichedUserMessage([...])` array over-indented by two spaces relative to the new call shape.

---

Findings 1 through 3 plus the indentation are applied and verified locally: `node --check` on all four files, `node test/run.js` 2193 passed / 0 failed, `selection-scope-restoration.mjs`, security corpus 60/60, provider-limits and toolbar-guard all green. Happy to push them onto the branch if you'd rather not hand-apply.

*Reviewed with Claude Code.*

esokullu:
Findings 1 through 3 and the indentation nit are now on the branch as `97470cb75` (fast-forward from `d7b60c764`, Chrome and Firefox in parity). Full suite against that exact tree: 2193 passed, 0 failed.

Leaving finding 4 for you to decide, but I traced the control flow properly afterwards and want to write down what I found, because it looks less like a design preference than it did on first read.

`_buildInitialUserMessage` reads the pending set and marks `enriched` at `:5605`. `_maybeRunPlannerGate` pushes the message and consumes the flag at `:14948`. The deterministic read does not run until `:33615`, and there are two early returns in between:

- `proceed:false` from the read-scope classifier returns at `:33378`
- `responseOnly:true` returns at `:33393` through `_completeResponseOnlyTurn`

Neither path reaches `:33615`, and the set is empty by then, so the user's next turn carries neither the runtime note nor the forced read. On the classifier path that means: user restores broader context, asks something, gets a clarifying question back, answers it, and that answer is handled with the pre-restore behaviour. Which is the case #2961 opened for.

The read-scope classifier is reachable here by construction, incidentally. It runs when `!runIntent && !sourceBoundRun`, and a restored turn is `!sourceBoundRun` by definition since the scope was just cleared.

`responseOnly:true` is the sharper of the two. The planner decides no tools are needed and the model answers from context, and on a just-restored turn that context is the selection-only history you are trying to get out from under.

What tips it for me is the asymmetry in what each mistake costs. Leaving the flag armed one turn longer than needed costs a redundant `read_page` on the current page, and a new selection, new conversation, independent run or tab teardown all clear it anyway. Consuming it early costs the feature working at all on those paths. The premise of this PR is that the prompt note by itself was not enough to stop the model following selection-only history, and the note by itself is exactly what those two paths fall back to.

The change I had in mind is small: consume at the point the turn commits to the tool loop rather than at message push, so move the `:14948` call to after `gateOutcome.proceed` and the `responseOnly` check have both cleared. It would need the regression test flipped, since the current one asserts the present timing.

I have not pushed any of that. It reverses a semantic you specified on purpose ("clear the restoration state after one use") and that your test pins down, so it is your call rather than mine. It also does not seem like a merge blocker to me; the ordinary path works and this reads like a follow-up commit.

Smaller instance of the same thing, spotted while tracing: the consume also fires on the dev-mode-blocked early return at `:33288` and `:34543`, where the turn ends before any model call happens at all. Low stakes since it is a dead end either way, but the restoration is spent there for a turn the user never really got.

esokullu:
Please move the consume call so the restoration state remains active until the turn has passed the proceed and responseOnly checks. The regression test should be updated to cover those paths as well.

16Miku:
Implemented the requested restoration-consumption change in 695fc37a. The pending state is now consumed only after both the proceed and responseOnly early-return checks have cleared, and it remains pending for read-scope clarification, response-only, and compact-provider Dev-blocked exits. The change is mirrored across Chrome and Firefox, streaming and non-streaming paths, and consumption is persisted when execution begins.\n\nVerification on the pushed commit:\n- node --check for both agent files\n- node test/selection-scope-restoration.mjs\n- npm test: 2193/2193 main tests and 60/60 security checks\n- Real Chrome trace: a responseOnly acknowledgment left restoration armed, and the next turn deterministically ran read_page before answering from the full page\n\nI also fetched the current upstream main before pushing; its newer changes are in unrelated UI/MCP/release areas and do not conflict with these three files.

PR #2970, opened 2026-09-03
fix(mcp): surface stable permission decisions

### LINKED ISSUE #2974: Add an explicit prompt kind for structured MCP gates
## Context

PR #2970 fixes localized approvals for `permission`, `submitConfirmation`, and `workflowHealing` prompts in the MCP and LM Studio bridges. See [PR #2970](https://github.com/webbrain-one/webbrain/pull/2970).

The current implementation infers the prompt type from whichever `pendingInput` field happens to be present. That works for the gates covered today, but a future structured gate could silently fall through to the free-text path. The client would then forward an answer that the extension rejects, and the user would only see the operation fail closed.

## Proposal

Add an explicit `promptKind` (or equivalent discriminator) to `pendingInput`. Have both bridges switch on it, and treat unknown kinds as structured or unsupported rather than free-form. This would make new gate types fail visibly instead of silently failing closed.

## Open design question

Where should answer normalization and validation live?

Today the bridge is advisory: it tells the client which exact values to send, but it does not enforce them. Meanwhile, `webbrain_respond` cannot validate the answer without knowing which gate is pending.

Should normalization and validation remain in the client, or should the protocol carry enough gate context for server-side validation? If server-side validation is preferred, what should the protocol shape be so that unknown or newly added structured gates fail safely and clearly?

This issue tracks that follow-up design work separately from the implementation in #2970.

16Miku:
## Summary

- distinguish structured permission prompts from ordinary free-text clarifications
- expose the stable `once`, `always`, and `deny` decisions to MCP clients
- guide clients to map an explicit normal approval to `once`, while reserving `always` for an explicitly persistent grant
- keep ordinary clarification answers verbatim and preserve fail-closed behavior for unknown permission values
- document the permission contract and add regression coverage

## Problem

The MCP tool previously told clients to forward every user answer verbatim. For a structured permission prompt, a localized approval such as Chinese “允许” could therefore be sent directly to the extension. The extension intentionally accepts only `once`, `always`, or `deny`, so an unknown value failed closed and the requested action was denied.

## Testing

- `npm run build`
- `node --test test/bridge.test.mjs test/catalog.test.mjs` — 13/13 passed
- manual local MCP + Chrome test: an explicit Chinese approval was mapped to `once`, WebBrain navigated the selected tab, and the final Chrome URL was independently verified

A cold-start test in a separate Codex conversation did not reach the permission step because the configured model provider returned empty completions without tool calls, so that scenario remains inconclusive rather than failed. Full-suite runs on this Windows machine also encountered timing-sensitive failures in existing timeout and stdio-shutdown tests under load; the changed test files pass cleanly.

### DIFF
--- a/lmstudio-plugin/src/index.ts
+++ b/lmstudio-plugin/src/index.ts
@@ -235,11 +235,20 @@ const browserRespondTool = tool({
   description:
     "Answer a question from a browser run whose status is needs_user_input, then wait " +
     "for the run to continue. The answer must come from the user; never invent it. Use " +
-    "the runId and clarifyId returned by browser_task or browser_status.",
+    "the runId and clarifyId returned by browser_task or browser_status. When that result " +
+    "carries a `decisions` list, the pause is a structured gate (permission, form submission, " +
+    "saved-workflow repair): send exactly one of those values, never a localized label, " +
+    "because the extension fails closed on an unknown decision.",
   parameters: {
     runId: z.string().min(1).describe("The browser run waiting for an answer."),
     clarifyId: z.string().min(1).describe("The pending clarification to answer."),
-    answer: z.string().min(1).describe("The user's answer, forwarded verbatim."),
+    answer: z
+      .string()
+      .min(1)
+      .describe(
+        "The user's answer, forwarded verbatim — or, when the pause reported a `decisions` " +
+          "list, exactly one of those stable values.",
+      ),
     timeout: z
       .number()
       .optional()
--- a/lmstudio-plugin/src/tools/browserTask.ts
+++ b/lmstudio-plugin/src/tools/browserTask.ts
@@ -29,6 +29,8 @@ export interface BrowserTaskResult {
   needsUserInput?: boolean;
   question?: string;
   clarifyId?: string;
+  /** Stable values a structured gate accepts. Absent for free-text clarifications. */
+  decisions?: string[];
   stillRunning?: boolean;
   finalUrl?: string;
   text?: string;
@@ -65,9 +67,38 @@ function bodyOf(snapshot: CloudSnapshot): string {
   return snapshot.content || snapshot.summary || "";
 }
 
+const isRecord = (value: unknown): value is Record<string, unknown> =>
+  value != null && typeof value === "object" && !Array.isArray(value);
+
+/**
+ * Permission, form-submission and workflow-repair pauses are structured gates,
+ * not questions: the extension accepts only the stable values it offers in
+ * `options` and fails closed on anything else, so a localized approval relayed
+ * verbatim ("允许") silently denies the action. Only the plain `clarify` tool
+ * takes free text.
+ */
+function decisionsOf(pending: Record<string, unknown>): string[] {
+  if (!isRecord(pending.permission) && !isRecord(pending.submitConfirmation)
+    && !isRecord(pending.workflowHealing)) {
+    return [];
+  }
+  const candidateIds = isRecord(pending.workflowHealing) && Array.isArray(pending.workflowHealing.candidates)
+    ? pending.workflowHealing.candidates
+      .map((candidate) => (isRecord(candidate) ? String(candidate.id ?? "").trim() : ""))
+      .filter((id) => id !== "")
+    : [];
+  const options = Array.isArray(pending.options)
+    ? pending.options
+      .filter((option): option is string => typeof option === "string" && option.trim() !== "")
+      .map((option) => option.trim())
+    : [];
+  return [...candidateIds, ...options];
+}
+
 function resultOf(snapshot: CloudSnapshot, timedOut = false): BrowserTaskResult {
   if (snapshot.status === "needs_user_input") {
     const pending = snapshot.pendingInput ?? {};
+    const decisions = decisionsOf(pending);
     return {
       ok: false,
       runId: snapshot.runId,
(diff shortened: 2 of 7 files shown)


esokullu:
Reviewed this and traced the contract through to the extension side (`src/chrome/src/agent/agent.js`, `src/chrome/src/cloud-runs.js`) to check what actually survives into `pendingInput`. The core change holds up: `_promptPermission` does emit `permission: { capability, host }` with `options: ['once','always','deny']`, and neither `cloudSafeUpdateData` nor `scrubCloudValue` strips it, so the new detection fires on real payloads rather than on a shape that only exists in the test.

The problem is the scope. Four findings, and the first two are the same bug this PR is fixing, still live on other gates.

### 1. Submit confirmations fall into the verbatim branch

`mcp-server/src/runs.ts:148`

The structured-prompt test looks only at `pendingInput.permission`, so `submitConfirmation` lands in the new `else`, which now states outright that the answer is forwarded verbatim. `_promptSubmitConfirmation` (agent.js:18329) emits `options: ['once','deny']` and accepts only `once|submit|confirm|allow|yes`, returning `'deny'` for anything else.

So a user approving a form submission in Chinese or Turkish gets that answer relayed as typed, it matches nothing, and the submission is blocked. Same failure mode as the one in your problem statement, on a gate reachable from any cloud or MCP run (agent.js:8472). The PR arguably makes this one slightly worse, since the free-text branch now tells the client explicitly to pass the text through.

### 2. Workflow-healing prompts cannot be answered at all

`mcp-server/src/runs.ts:148`

`workflowHealing` (agent.js:18367) requires an answer matching `^candidate_(\d)$`, and the candidate ids sit in `pendingInput.workflowHealing.candidates[].id`. `describeSnapshot` surfaces neither the ids nor the allowed values, so there is nothing in the snapshot an MCP client could use to construct a valid answer. Whatever it sends fails the regex and `_promptWorkflowTargetHealing` returns `null`, declining the repair without saying why.

### 3. Decisions are hardcoded rather than read from the payload

`mcp-server/src/runs.ts:151`

`permission_decisions: once | always | deny` is a literal, but the extension already sends `pendingInput.options`. If the emitter ever narrows the set for a sensitive capability, the MCP text goes on advertising a value the gate no longer accepts.

### 4. The LM Studio bridge still has the original bug

`lmstudio-plugin/src/tools/browserTask.ts:78`

The sibling bridge builds the same `needs_user_input` hint ("Ask the user this question, then call browser_respond") with no permission guidance, and `browser_respond` documents `answer` as forwarded verbatim. The localized-approval bug is untouched on that surface, so fixing only the MCP path leaves the same report reproducible through LM Studio.

---

All four are fixed and pushed to this branch as `a7fc14717`.

The pause now branches on prompt kind, with a separate structured branch for `permission`, `submitConfirmation` and `workflowHealing` plus the free-text fallback. Each structured branch prints its accepted values and asks for exact-value mapping, and workflow healing additionally prints each candidate id with its target.

One judgement call worth flagging, since it shaped the fix: it keys on prompt kind rather than on the mere presence of `options`. The plain `clarify` tool (agent.js:25936) also ships `options` while genuinely accepting free text, so treating "has options" as "exact values only" would have regressed ordinary clarifications.

Also in the commit: `mcp-server/src/index.ts` and the README carry the `decisions:` line for the non-permission gates, two regression tests in `bridge.test.mjs` follow the style of your permission test, and the LM Studio side gets the parallel change in `browserTask.ts` and `index.ts`.

What I deliberately left alone: no server-side normalization or validation of `answer`. `webbrain_respond` has no snapshot at call time, so it cannot know which gate is pending, and the extension's fail-closed handling looks intentional. That is a design decision for you rather than something to slip into a fix. There is also no regression test on the LM Studio side; its harness replies with one fixed clarify payload and `resultOf` is not exported, so covering it properly means reshaping the mock.

Verification on the pushed tree: `mcp-server` builds clean, 30/30 tests pass, up from 28 with the two new ones. `lmstudio-plugin` builds clean, 12/12. Worth noting for your Windows runs that neither package had `node_modules` here, so the `tsc` errors on a fresh clone are missing deps rather than anything wrong with the branch.

*Reviewed with Claude Code.*

esokullu:
One structural point I described only obliquely in the review, worth stating directly because it is the reason this bug reached review twice.

`describeSnapshot` works out which gate is pending by checking which key happens to be present: `permission`, then `submitConfirmation`, then `workflowHealing`. I went looking for a discriminator on `pendingInput` and there is none — no `promptKind`, no `exactValues`, nothing that names the gate type. The kind is inferred from the shape.

That inference is what let the original version ship. It handled `permission`, the other two structured gates fell through to the free-text branch, and nothing anywhere reported it. My commit takes the count from one gate to three, but it does not make the fourth safe. The next structured gate added on the extension side lands in the same fallback, has its answers forwarded verbatim, fails closed at the gate, and the only signal is a user reporting that approvals do not work.

The default is also the wrong way round for safety. An unrecognized pause is currently treated as free text, which is the permissive reading. Treating it as exact-value-only would fail loudly on a gate the bridge does not understand yet, instead of quietly relaying text that will be rejected downstream.

A `promptKind` on `pendingInput`, with both bridges switching on it and treating an unknown kind as structured rather than free-form, would move this from something reviewers have to notice to something the type system carries. It touches the extension, the MCP server and the LM Studio plugin together, so it is a maintainer call and clearly outside this branch. Filing it here so it is written down rather than rediscovered on the next gate.

The other thing left open from the review, restated for the same reason: the fix is advisory end to end. `describeSnapshot` prints the accepted values and asks for an exact one, but nothing verifies the client complied, so a client that ignores the instruction still reaches the gate with text that fails closed. Server-side validation would need `webbrain_respond` to know which gate is pending, which it cannot today, so that is a design decision about where normalization belongs rather than a fix.

16Miku:
Thank you for the thorough review and for tracing the payload all the way through the extension. I have fast-forwarded my local branch to `a7fc14717` and reviewed the additional changes. The submit-confirmation and workflow-healing findings make sense, as does reading the accepted values from `pendingInput.options` instead of hardcoding them.

Local verification on the updated branch:

- `mcp-server`: build passes; 30/30 tests pass
- `lmstudio-plugin`: build passes; 12/12 tests pass
- Real MCP + Chrome navigation test: a Chinese one-time approval (`允许`) was mapped to `once`, and the selected tab navigated to `https://example.com/`; the final Chrome URL and title were independently verified
- Real local form test: the type permission was approved with `once`, then the new submit-confirmation branch surfaced `decisions: once | deny`; a Chinese approval was mapped to `once`, and the POST completed on a loopback-only test server. The resulting Chrome URL was independently verified as `/submitted`

I did not manually exercise workflow healing or the LM Studio UI end to end; those paths were covered by their package tests here.

I also agree with the structural point about adding an explicit `promptKind` discriminator and making unknown structured gates fail loudly. That looks better as a separate cross-surface design change rather than expanding this fix further.

Thanks again for strengthening the patch.

PR #2978, opened 2026-09-04
fix: defer broader-context restoration consumption past proceed and r…

webbrain-one:
…esponseOnly checks


webbrain-one:
@copilot resolve the merge conflicts in this pull request

esokullu:
Closing this as superseded by #2968.

That PR (695fc37a1) landed the same fix: on `main`, `src/chrome/src/agent/agent.js:33448` already runs the consume after the `responseOnly` early return. The behavioral coverage is there too, at `test/selection-scope-restoration.mjs:168` and `:263`, which drive the early-return paths end to end across chrome and firefox, streaming and non-streaming, for the blocked, response-only, and dev-blocked cases.

What's left unique to this branch is three source-text regex assertions, two gate-level test blocks, and a brace reformat in both agent bundles. The regexes grep the source for exact formatting rather than testing behavior, and I checked by mutation that they don't catch anything the existing tests miss. The two gate blocks duplicate the coverage above them. The reformat was only there to satisfy one of the greps.

That leaves a three-line explanatory comment, which isn't worth keeping the PR open for. Reopen if I've misread something.

Issue #2976, opened 2026-09-04
feat: show WebBrain actions for selected text in browser-rendered PDFs


alectimison-maker:
## Problem

When a user opens an online PDF directly in Chrome and selects text, WebBrain's floating selection actions do not appear. This affects browser-rendered PDFs from any source, including arXiv, publisher websites, documentation sites, and CDNs.

The normal web-page experience is implemented by content scripts listening for selection changes and positioning the floating action near the selected range. Chrome's built-in PDF viewer is a separate privileged viewer surface, so the existing content script cannot reach its internal text layer.

This should be solved as a browser-wide PDF capability rather than as a site-specific integration.

## Goal

Enable WebBrain actions for text selected in online PDFs viewed in Chrome, with behavior consistent with the existing web-page selection experience:

- Show the WebBrain floating action after text is selected.
- Reuse the existing selection prompt, popup, side-panel handoff, localization, and untrusted-content handling.
- Preserve the original PDF URL and tab scope when submitting the selected text.
- Keep a reliable fallback for PDFs that cannot be handled by WebBrain.

## Proposed direction

For Chrome 151+, use the public `chrome.mimeHandler` API to handle `application/pdf` responses in a WebBrain extension page:

1. Receive Chrome's one-time PDF stream through `chrome.mimeHandler.getStreamInfo()`.
2. Render the document with the vendored PDF.js bundle.
3. Render PDF.js's text layer over the canvas so the browser Selection API can provide selected text and client rectangles.
4. Reuse or extract a shared selection controller for the existing web-page and PDF surfaces.
5. Fall back to Chrome's native PDF viewer when the stream is unavailable or the document cannot be rendered.

The implementation should not depend on private Chrome PDF viewer internals, debugger attachment, or injected scripts inside the built-in viewer.

## Delivery plan

All items below are part of the intended feature. They are ordered so that each stage has a useful, testable result before the next one begins.

### Phase 0: Prove the selection path

- [ ] Chrome 151+ top-level online `application/pdf`.
- [ ] PDFs from multiple hosts, not only one provider.
- [ ] Single-page canvas plus selectable PDF.js text layer.
- [ ] Selection-change handling, range positioning, and the WebBrain floating action.
- [ ] Submission through the existing background, popup, and side-panel path.
- [ ] Native fallback for loading or rendering failures.

### Phase 1: Complete top-level PDF support

- [ ] Multi-page rendering, scrolling, page navigation, zoom, rotation, and browser zoom.
[...]

PR #2980, opened 2026-09-04
feat: show WebBrain actions for selected text in online PDFs

alectimison-maker:
## Summary

Implements the online-PDF selection action flow described in #2976.

WebBrain can now render browser-handled PDFs in an extension-owned PDF.js viewer with a selectable text layer, so the existing selection shortcut appears for PDF text as it does on normal pages.

Refs #2976

## What changed

- Adds a Chrome `application/pdf` MIME handler for Chrome 151+.
- Reads Chrome's one-time PDF stream and renders canvas plus PDF.js text layers.
- Adds page navigation, zoom, fit width, rotation, search, download, and print controls.
- Supports embedded PDFs through `can_embed` and keeps selection scoped to the PDF handler tab.
- Adds a bounded, manual OCR fallback for scanned pages through the configured vision route.
- Adds an explicit WebBrain PDF viewer context-menu entry for Chrome and Firefox.
- Adds the Firefox explicit viewer fallback and mirrors the selection/OCR/background contracts.
- Adds a Chrome settings opt-in (`pdfViewerEnabled`, off by default); explicit viewer links remain available when the global MIME handler is disabled.
- Falls back to the native viewer for disabled, unavailable, unreadable, or unrenderable MIME-handler flows.

## Validation

- `npm run test:pdf-selection` — 9 passed.
- `node test/run.js` — 2194 passed; 2 pre-existing repository failures remain:
  - changelog newest entry is `34.1.4` while `package.json` is `34.1.5`;
  - tracked `dist/webbrain-chrome-34.1.5.zip` is missing.
- `git diff --check` and JavaScript syntax checks passed.
- Headed Google Chrome 153 verification covered native fallback with the opt-in off, automatic WebBrain routing with it on, explicit viewer loading, multi-page rendering, selection handoff, embedded PDF loading, OCR fallback state, navigation, zoom, rotation, search, download, and print controls.
- Firefox manifest, handler, OCR, context-menu, and tab-scoped selection contracts were verified statically; Firefox runtime automation is not available in the repository test harness.

## Implementation notes

- PDF text and OCR output remain untrusted page data and use the existing selection-only prompt/security path.
- The handler consumes `streamUrl` once and does not re-fetch the original URL for MIME-handler navigations.
- The implementation work is preserved as separate reviewable commits, followed by a focused hardening commit for review findings.


webbrain-one:
cursor comments:

Chrome gesture break. WB_PDF_SELECTION_SHORTCUT_SUBMIT awaits tabs.get before sidePanel.open, so the panel often won’t open (unlike the sync normal selection path). Open/queue first; resolve the tab after.
Large-PDF OOM. renderAllPages mounts every page’s canvas + text layer up front. Virtualize before merge.
Weak tab-scope. Only “tab exists”; any extension sender can pass msg.tabId. originalUrl is unused in Chrome prompt building. Verify sender.url is the PDF handler.
Over-broad context menu. *://*/*.pdf* matches non-PDFs that contain .pdf, then re-fetches with credentials. Tighten pattern / require real PDF type.
OCR and MIME fallback look fine. Firefox’s requiresManualOpen: true is the safer gesture story.

webbrain-one:
@copilot resolve the merge conflicts in this pull request

webbrain-one:
@copilot resolve the merge conflicts in this pull request

esokullu:
## Review — WebBrain actions for selected text in online PDFs

**Verdict: Approve with nits.** The security architecture is sound (sender scoping, tab liveness checks, untrusted-data path, bounded OCR), the Chrome/Firefox split is well-reasoned and documented, and the full suite is green after the merge (`test/run.js` 2198 passed, `test/pdf-selection.mjs` 9 passed). No blocking issues. Findings below are ordered by importance; none should hold up merge.

### Strengths

- **Tab-scope can't be forged cheaply.** `isPdfHandlerSender` (extension-id + viewer-URL prefix + `tabId` match) plus `tabs.get(tabId)` liveness before `queueSelectionShortcutPrompt` is the right shape for both selection submit and OCR. Chrome preserves the sidePanel gesture by opening synchronously before the async tab lookup — correct and clearly commented.
- **OCR distrust is done properly.** System prompt refuses image instructions, `_wrapUntrusted('pdf_ocr_image', …)` keeps it in the selection-only path, boxes are normalized/clipped/bounded (300 lines, 500 chars/line, 20k total), rendering uses `textContent`, and the trace deliberately omits recognized text. 16 MB image cap + vision-timeout mapping + user cancel via `AbortController` round it out.
- **OOM fix is real.** Eager ≤20 pages, placeholder shells + windowed render with far-page eviction, `renderSequence` cancellation, pixelRatio capped at 2. Sensible 64 MB / 500-page guardrails with native fallback.
- **Opt-in default is safe.** Chrome auto-routing gated behind `pdfViewerEnabled === true` (off by default); explicit viewer always available. Firefox keeps native default with explicit-only entry, documented in `ARCHITECTURE.md` including the POST-navigation limitation. Honest scoping.
- **Graceful degradation everywhere.** Every viewer control failure funnels to `fallbackToNative`, and `mimeHandler.getStreamInfo` absence / disabled flag / oversize all fall back instead of stranding the user.

### Findings (all non-blocking)

1. **Nit — copy/paste string in Firefox handler.** `src/firefox/src/ui/pdf-handler.js` `loadPdfBytes` path throws `'Chrome returned an empty PDF stream.'` — should say Firefox. (Its sibling message correctly says "Firefox's native viewer".)
2. **Consider — `fetch_pdf_document` is a credentialed fetch proxy.** It validates http/https + sender + tab liveness, but not that the URL matches the claiming tab or is actually a PDF. Reachable only from extension-origin viewer pages and results never leave the extension, so risk is low — but a content-type check or URL-vs-tab consistency check would tighten it.
3. **Consider — `pdfResponseTabs` staleness.** Cleared on non-PDF `main_frame` headers and tab close, but an SPA-style navigation that doesn't produce headers could leave a stale entry. Worst case is a visible-but-no-op menu item (`handleContextMenuAsk` re-validates), so acceptable; noting for awareness.
4. **Consider — scroll handler is O(n) per event** (`updateCurrentPageFromScroll` does `getBoundingClientRect` over up to 500 shells, unthrottled). Passive listener and no forced render, so fine in practice; throttle if jank is ever reported on 500-pagers.
5. **Consider — detached `anchor.click()` for download** works under user gesture in Chrome/Firefox, but appending to DOM first is the more bulletproof pattern if any download issue is ever reported.
6. **Test gap (acknowledged by author).** Firefox runtime paths are static-assertion only; no real-PDF render or OCR integration test (normalization is unit-tested, which is the important part). Consistent with repo conventions; fine.
7. **CI note.** `smoke` (WebMCP e2e) fails on `Timed out waiting for fixture page error` — unrelated to this diff (no WebMCP files touched); looks flaky/pre-existing. `dist/*` zips untouched, matching the PR's noted pre-existing release-state failure; follow normal release flow if a rebuild is required.

esokullu:
Addressed the non-blocking findings from the review. Pushed as commit `053ec9a04` (`fix: address PDF viewer review nits`).

**Finding 1 — copy/paste string.** The Firefox handler now throws `'Firefox returned an empty PDF stream.'` instead of the copied `'Chrome returned…'` message (src/firefox/src/ui/pdf-handler.js).

**Finding 2 — credentialed fetch proxy.** `fetchPdfDocumentForViewer` (src/firefox/src/background.js) now requires the proxied response to carry an `application/pdf` content type, so the credentialed fetch cannot be pointed at arbitrary HTML the viewer did not open. This uses the content-type branch of the reviewer's suggested tightening; the sender is already scoped to the viewer page carrying the matching tab id.

**Finding 4 — scroll scan cost.** `updateCurrentPageFromScroll` in both Chrome and Firefox handlers is throttled with `requestAnimationFrame` so the O(n) `getBoundingClientRect` scan runs at most once per frame instead of on every scroll event.

**Finding 5 — download reliability.** `downloadPdf` now appends the `<a>` to the document (hidden) before `.click()` and removes it after, the more bulletproof user-gesture download pattern.

**Tests.** Added `testPdfHandlerHardeningReviewFindings` covering all four; `test:pdf-selection` passes 10/10. Findings 3 (SPA staleness), 6 (Firefox runtime test gap), and 7 (unrelated CI flake) were left as-is per the review.

`git diff --check` and per-file `node --check` are clean.

esokullu:
Resolved all eight PDF viewer review findings in d92a0f3df.

Fixes:
- Downloads now request fresh bytes from the loaded PDFDocumentProxy, avoiding the ArrayBuffer that PDF.js transfers to its worker and avoiding a persistent second 64 MiB copy.
- Chrome stream loading and the Firefox background fetch proxy now read incrementally and cancel immediately when the 64 MiB cap is crossed, including absent, malformed, or understated Content-Length cases.
- Printing a virtualized PDF now renders every page before opening print, then restores the prior page and virtualized window afterward.
- Explicit top-level viewer failures restore the original tab URL before attempting the MIME-child-only fallback API.
- Manual scrolling synchronizes requestedPage so zoom, fit, rotate, and resize retain the visible page.
- Each virtual render window has its own token, so a stale pass cannot release the current window after rapid scrolling.
- Chrome creates the PDF menu item hidden and synchronizes visibility through supported active-tab, tab-update, window-focus, and response-header hooks; the nonexistent contextMenus.onShown hook is removed.
- Firefox creates the item hidden and calls menus refresh after each onShown visibility update.

Additional finding fixed during implementation:
- Small PDFs were also scheduling the large-document render window after scrolling, which could discard eagerly rendered pages outside the window. Window scheduling and page release are now restricted to PDFs above the eager-page threshold.

Regression coverage:
- Added executable stream-limit tests covering an understated length, early cancellation, malformed length, and chunk reconstruction.
- Added Chrome/Firefox assertions for fresh download bytes, complete print rendering and re-virtualization, scroll target retention, stale-window tokenization, supported Chrome menu hooks, Firefox refresh, and the small-PDF guard.
- npm test passes: 2,198 core tests, 11 PDF tests, 33 toolbar-guard tests, selection-scope restoration, offline relevance benchmark, and 60/60 security checks.

esokullu:
Hosted-check follow-up:
- The WebMCP smoke job on d92a0f3df failed at test/webmcp-e2e.mjs:133 while waiting for a fixture page error, with received: []. The preceding 053ec9a04 run failed at the same assertion before these PDF fixes, so this is a pre-existing WebMCP smoke failure rather than a PDF regression.
- Current run: https://github.com/webbrain-one/webbrain/actions/runs/33894215344
- Pre-fix run with the identical failure: https://github.com/webbrain-one/webbrain/actions/runs/33889924699
- Vercel also reports its repository authorization failure on this PR SHA.

The complete local npm test suite remains green, including 2,198 core tests and all new PDF regressions.

esokullu:
## Code review (high effort), fixes applied in 89604b6

Reviewed the 25-file diff. Nine issues found, eight fixed and pushed; two left for follow-up with reasons below. `npm run test:pdf-selection` passes 11/11 and `node test/run.js` passes 2198/0 after the fixes.

### Fixed

**Page rotation is dropped** (`src/chrome/src/ui/pdf-handler.js:218`, also `:150`, `:219`, `:333`, and the Firefox copy)

`page.getViewport({ rotation: state.rotation })` overrides PDF.js's `rotation = this.rotate` default, which throws away the page's own `/Rotate` entry. A landscape scan or fax stored as `/Rotate 90` renders sideways in the WebBrain viewer while Chrome's built-in viewer shows it upright, and `renderScaleFor` then fit-widths against the wrong axis. Now composed via `pageRotation(page)` so the viewer rotation stacks on top of `page.rotate`.

**Text layer inherits the body font** (`src/chrome/src/ui/pdf-handler.css:159`, and the Firefox copy)

The vendored PDF.js 5.7 `TextLayer` stopped setting an inline `font-size`. It publishes `--font-height`, `--scale-x`, `--rotate`, `--min-font-size` and `data-main-rotation`, and leaves the embedding stylesheet to turn those into a real size. `pdf-handler.css` styled none of them and never defined `--total-scale-factor` or `--scale-round-*`, so every selectable span rendered at the 14px body size with no horizontal stretch. Selecting a 28pt heading or a 6pt footnote highlighted a band that did not match the glyphs on the canvas, which is the thing this feature exists to do. Added the span `font-size`/`transform` rules, the `[data-main-rotation]` counter-rotation rules, `--scale-round-x/y`, and an `applyTextLayerScale()` that sets `--total-scale-factor` from `viewport.scale * viewport.userUnit`.

**Content-type tracking is cleared before it can be used** (`src/chrome/src/background.js:1771`)

`chrome.tabs.onUpdated` deletes the `pdfResponseTabs` entry when `changeInfo.url` fires, but `webRequest.onHeadersReceived` records that entry while the response is still in flight, and the tab URL only changes at commit, so afterwards. Navigating to something like `https://host/download?id=7` served as `application/pdf` marks the tab and then unmarks it a moment later, so "Open PDF with WebBrain" never appears for a PDF whose path has no `.pdf`, which is the case the tracking was added for. Firefox has no such deletion and behaves correctly. Removed the delete: `trackPdfResponse` already replaces the entry on the next main-frame response, and `onRemoved` covers tab close.

**Empty `markedContent` wrappers count as text** (`src/chrome/src/ui/pdf-handler.js:106` and `:246`)

`querySelector('span')` and `span:not([data-webbrain-ocr])` also match PDF.js's empty `span.markedContent` wrappers, which tagged PDFs emit on image-only pages. On a scanned page that makes `currentPageHasNativeText()` return true, so the "OCR page" button never shows up, and it inflates `state.textLayerCount` enough to suppress the "no selectable text" warning. Added `hasNativeTextSpans()`, which requires a span holding non-empty text.

**Current page flips early on tall pages** (`src/chrome/src/ui/pdf-handler.js:384`)

`updateCurrentPageFromScroll` picked the page whose top edge sat nearest the stage top. On a page taller than the viewport the next page's top becomes the nearer one past the midpoint, so the counter moves to page N+1 while you are still reading page N. `ocrCurrentPage()` reads `state.currentPage`, so "OCR page" would OCR and render text onto the wrong page. Now picks the page with the largest visible intersection, falling back to the nearest-top scan.

**OCR spans have a box but no font size** (`src/chrome/src/agent/pdf-ocr.js`, both browsers)

OCR spans get explicit px `width` and `height` but no `font-size`, so a 14px glyph run sits inside a box that may be 30px tall and only a thin band of each line is selectable. Now sets `span.style.fontSize` from the reported box height.

**Stale PDF marker in Firefox** (`src/firefox/src/background.js:254`)

`trackPdfResponse` has `if (!contentType) return;` in Firefox but not in Chrome, so a main-frame response with no `content-type` header leaves the previous PDF marker on the tab and the menu keeps offering the viewer for a page that is no longer a PDF. Dropped the guard for parity with Chrome, where `String(undefined)` fails the regex and clears the entry.

**Cleanup**

`state.pageSizes` was written in three places and read nowhere, so it is gone. Added the missing `[data-kind="success"]` status rule (used in four call sites) and a rule for the `canvas.hiddenCanvasElement` that PDF.js appends to `<body>` for text measurement.

### Left for follow-up

**Credentialed fetch with no content-type check** (`src/chrome/src/ui/pdf-handler.js:668`)

The explicit-viewer path fetches `?url=` with `credentials: 'include'` and never checks the response content type, unlike Firefox's `fetchPdfDocumentForViewer`, which requires `application/pdf`. A crafted `chrome-extension://…/pdf-handler.html?url=…&tabId=N` link gets the extension to issue a credentialed cross-origin request to an arbitrary http(s) URL. Nothing reaches the page, since PDF.js just fails to parse the response, so the impact is limited to the request itself. Adding the Firefox check here would reject PDFs served as `application/octet-stream`, which is a behavior change beyond what this diff intends, so it seemed better as its own change.

**Untranslated menu title** (`src/chrome/src/background.js:449`, and the Firefox copy)

`"Open PDF with WebBrain"` is hardcoded English while every sibling item comes from `getSelectionShortcutLocalization()`, so non-English users see one untranslated entry. Fixing it needs a new key added across the locale dictionaries.

---

The Chrome and Firefox copies of `pdf-ocr.js`, `pdf-stream.js`, and `pdf-handler.css` are still byte-identical, which `test/pdf-selection.mjs` asserts for the stream module.

https://claude.ai/code/session_01UPwwLYz563fNemyqbmAHJT

esokullu:
## Merge assessment

Read the whole diff, not just the parts the review pass touched. Short version: net plus, and I would merge it, but there is one question I cannot answer from the code that should be settled first.

### Why it is a net plus

Selectable text on online PDFs, with WebBrain actions on that selection, closes a real gap, and the OCR path extends it to scans. The design is conservative where it counts. MIME interception is off by default (`pdfViewerEnabled === true`, so unset means off). The OCR call goes through the existing vision route with the untrusted-content wrapper, and it deliberately keeps recognized text out of the trace so a metadata-only trace cannot turn into a copy of the document. The 64MB and 500-page ceilings bound the worst case.

Blast radius on existing users is small. `agent.js` is additive: one new method, plus an optional parameter on `_withVisionDeadline` that is a no-op when null. `selection-shortcut.js` reads its config from the isolated-world global, which page content cannot set, and behaves as before when that config is absent.

### Settle this before merging

`src/chrome/manifest.json:33` registers `mime_types_handler`. That key is restricted to allowlisted and component extensions on shipping Chrome, and the settings copy says "On Chrome 151+", so this targets a capability I cannot exercise here. Two cases:

1. **Chrome ignores the key today.** Then the toggle does nothing for every current user, and a visible setting ships with no effect. Worth confirming the key does not draw a manifest warning or a store-review flag either.

2. **Chrome honors the key.** Then the handler page loads on every PDF navigation for every user, including those who never enabled the toggle. Those users reach `initialize()`, fail the gate, and land in `fallbackToNative()` with an empty message, which calls `abortAndFallbackToNativeHandler()`. If that API is missing, named differently, or throws, the catch is silent and the page sits on the `"Loading PDF…"` string from `pdf-handler.html:44`. Someone who never opted in would see a blank-looking page where their PDF used to open.

The second case is the only way this PR breaks something that currently works, and it lands on people who did not opt in. The fix is small: in the disabled path, if the abort fails, navigate the tab to the original URL, or at minimum surface an error rather than leaving the loading string up. Worth doing regardless of how the Chrome question resolves.

### Other places bugs could come from

**A tradeoff in one of the fixes just pushed.** Removing the `tabs.onUpdated` delete fixed the case where the menu item never appeared, but `pdfResponseTabs` now clears only when a new main-frame response arrives. Back/forward navigations restored from bfcache do not fire `webRequest`, so an entry can go stale and the menu item can linger on a page that is no longer a PDF. Clicking it navigates to the viewer, which fails and bounces back, so it recovers on its own, but it is ugly. The fully correct version clears on `changeInfo.url` only when the committed URL differs from the one recorded.

**Service worker churn.** `syncPdfContextMenuForActiveTab()` runs on every main-frame `onHeadersReceived` and on every `tabs.onUpdated` carrying `url` or `status`, each time doing a `tabs.query` plus a `contextMenus.update`. Several no-op round trips per navigation per tab. Not a correctness problem, but on a tab-heavy user it keeps the worker awake more than it needs to, and it is cheap to gate on whether visibility actually changed.

**Test confidence is thinner than 11/11 suggests.** Most of the 11 tests are `assert.match` against source text: 83 pattern assertions over 29 file reads. They would catch a deleted guard. They would not have caught either High finding from the review, since both were about how this code interacts with PDF.js at runtime, and only the selection-shortcut tests drive a real browser. A rendering test that loads a `/Rotate 90` PDF and checks a text span's computed font size would be worth more than the rest combined.

**Low severity, for completeness.** The `WB_PDF_SELECTION_SHORTCUT_SUBMIT` listener opens the side panel for the claimed `tabId` before verifying the tab exists, and `isPdfHandlerSender` cross-checks the tab id only when the sender URL carries one, which MIME-handler loads do not. Only our own extension pages pass the sender check, so web content cannot reach it. Worth a comment rather than a blocker.

### Recommendation

Merge after two things: someone confirms `mime_types_handler` behavior on the Chrome versions this actually ships to, and the disabled-path fallback stops failing silently. Both are small. The bfcache staleness, the sync churn, and the two items left open in the review comment above are reasonable follow-ups.

https://claude.ai/code/session_01UPwwLYz563fNemyqbmAHJT

esokullu:
Follow-up after checking the current Chrome contract and exercising exact head 89604b694 in Google Chrome 152.0.7977.77:

The merge-assessment conclusion is still right, but the central diagnosis should be corrected. The dict-form mime_types_handler entry is public for application/pdf on Chrome 151+, not restricted to allowlisted/component extensions. On a fresh unpacked install, chrome.mimeHandler exposed getStreamInfo, abortAndFallbackToNativeHandler, getMimeHandlerOptions, and setMimeHandlerOptions. Navigating to a real PDF entered the WebBrain handler and successfully returned to the native viewer while the WebBrain setting was off.

That exposed the actual blocker: Chrome reports the native application/pdf handler option as enabled by default, while storage.local has no pdfViewerEnabled value and the WebBrain checkbox renders unchecked. settings.js only writes storage.local and never calls setMimeHandlerOptions, so “off by default” still intercepts every PDF and relies on abort-and-fallback.

I would merge after synchronizing pdfViewerEnabled with chrome.mimeHandler.setMimeHandlerOptions(application/pdf, { enabled }), explicitly initializing the native option to disabled when the stored WebBrain setting is absent, retaining the handler-side storage check as defense in depth, and adding a real-Chrome regression for native option state/routing. The silent abort failure can still be hardened, but replaying originalUrl is not a general MIME fallback because POST and single-use requests may not be repeatable.

The bfcache staleness and test-confidence observations remain valid. The service-worker churn and sender/tab note remain non-blocking.

esokullu:
I love it - thanks @alectimison-maker

PR #2989, opened 2026-09-07
fix: harden large tab-chat image persistence

alectimison-maker:
## Summary

- Replace the large-payload regular expression in tab-chat image compaction with a linear scanner.
- Preserve existing image data URL behavior, including MIME parameters and mixed-case markers.
- Keep Chrome and Firefox implementations aligned and add a 6 MB regression case.

## Context

The latest upstream main still contained the implementation discussed in PR #2982. The reported Maximum call stack size exceeded was intermittent: the original code passed two full runs and 20 isolated reproductions in this environment. This change removes the risky large-payload regex path and makes the reported 6 MB scenario explicit in the test suite.

## Validation

- npm test: 2202 passed, 0 failed
- Security corpus: 60/60
- Provider model limits: passed
- Rich-text toolbar guard: 33 passed

### DIFF
--- a/src/chrome/src/ui/tab-chat-persistence.js
+++ b/src/chrome/src/ui/tab-chat-persistence.js
@@ -6,11 +6,94 @@ const TAB_CHAT_QUOTA_RETRY_BUDGET = 256 * 1024;
 export const TRANSPARENT_PIXEL_PNG_DATA_URL =
   'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNkYPhfDwAChwGA60e6kgAAAABJRU5ErkJggg==';
 
+const IMAGE_DATA_URL_PREFIX = 'data:image/';
+const BASE64_DATA_URL_MARKER = ';base64,';
+
+function asciiLowerCode(code) {
+  return code >= 65 && code <= 90 ? code + 32 : code;
+}
+
+function matchesAsciiIgnoreCase(source, offset, expected) {
+  if (offset + expected.length > source.length) return false;
+  for (let i = 0; i < expected.length; i++) {
+    if (asciiLowerCode(source.charCodeAt(offset + i))
+      !== asciiLowerCode(expected.charCodeAt(i))) return false;
+  }
+  return true;
+}
+
+function isMimeTypeCode(code) {
+  return (code >= 48 && code <= 57)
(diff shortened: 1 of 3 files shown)


esokullu:
Reviewed this and pushed one fix to the branch (f390456d). The new scanner diverges from the old regex in a way that can delete transcript content.

`imageDataPayloadEnd` walked forward from `data:image/<mime>` hunting for `;base64,`, stopping only at a comma or the end of the string. The old regex required the metadata section to start with `;` right after the MIME type, so a bare `data:image/png` occurrence never matched at that position: someone typing the MIME type in a message, a code block that mentions it, an `` with no payload. The scanner instead skips the quote, the `>`, whole tags, whatever is there, until it finds a real `;base64,` later in the document. Everything from the first marker through that later payload is then replaced with the transparent pixel:

```
input:  <div>use data:image/png here</div><img src="data:image/gif;base64,AAAA">
old:    <div>use data:image/png here</div><img src="data:image/gif;base64,iVBOR...">
new:    <div>use data:image/png;base64,iVBOR...">
```

The `</div>` and the `` tag are gone, and that malformed HTML is what gets restored into the DOM on the next tab chat restore.

The same missing bound works against the performance goal here: every unmatched `data:image/` marker scans to the end of the string, so k of them in a 7 MB transcript cost O(k*n).

The fix requires the parameter section to begin with `;` and consumes each `;name=value` with a media type parameter character class. Any character that cannot belong to a parameter (a quote, an angle bracket, whitespace, a comma) ends the walk and rejects the match, so it can never run past the attribute it started in. Both the Chrome and Firefox copies were identical and both are patched.

Behavior now matches the old regex on everything I checked: bare mention then image, non base64 SVG, two normal images, mixed case with a `charset` parameter, empty payload, empty input, parameters containing `_` and `*`. One difference is deliberate. `data:image/png;charset=x"><b>hi</b><img src="data:image/jpeg;base64,...` also swallowed the intervening markup under the old regex, and the bound fixes that too.

Added a regression test for the bare mention case, run against both builds. Full suite is 2203 passed, 0 failed, including the test that came with this PR. A 6 MB payload with a preceding bare mention strips in about 17 ms.

Nothing else in the diff looked wrong to me. The `copyCursor`/`searchCursor` bookkeeping, the character classes, the case insensitive compare, and the `replaced` fast path all check out.

🤖 Review by [Claude Code](https://claude.com/claude-code)

PR #3033, opened 2026-09-13
fix(opencode): send x-opencode-session for OpenCode Go

### LINKED ISSUE #3031: OpenCode Go needs x-opencode-session header
Trying to use OpenCode Go results in the following error from OpenCode:
```
Failed: opencode-go error 400: {"type":"error","error":{"type":"MissingSessionID","message":"Error from provider (Console Go): Request is missing x-opencode-session and cannot be routed efficiently. Please see https://opencode.ai/docs/go/#where-can-i-use-it"}}
```

paulswartz:
OpenCode Go rejects chat requests that omit x-opencode-session. Forward the active per-tab conversation id as the session id, with a stable per-provider fallback for calls without a conversation (for example Test connection). Gate the header to the opencode-go provider in both Chrome and Firefox builds.

Tested the Firefox and Chrome version locally: Test Connection works, and the sidebar passed a smoke test (responding to a simple question)

Closes #3031


esokullu:
thank you! 🙏

PR #3062, opened 2026-09-17
fix: group low-risk search permissions

### LINKED ISSUE #3060: Group low-risk search actions into one permission request while preserving high-risk confirmation
## Problem

A simple Google Scholar search can require multiple permission or confirmation prompts:

1. Navigate to the site
2. Type the search query
3. Confirm the form submission

This makes a low-risk search feel overly interruptive and requires too much manual input.

The current permission model correctly separates capabilities such as navigation, typing, and clicking. Form submission also has a fresh confirmation step. However, these checks are currently evaluated per browser tool call rather than against the user's complete intent. As a result, one read-oriented search can produce several prompts.

GET search forms are also treated as submit actions when the clicked control belongs to a form, without a risk classification based on the HTTP method or field sensitivity.

## Proposed behavior

Classify the complete user intent before executing browser actions:

- For a same-site, read-oriented GET search with non-sensitive fields:
  - Ask for permission at most once.
  - One approval may cover the related navigation, typing, and search-control click.
  - Scope the approval to the current task and host; do not create a broad global grant.

- For high-risk actions such as changing passwords or account data, payments, deletion, publishing, or POST/PUT/PATCH/DELETE requests:
  - Keep a confirmation immediately before the final mutation.
  - Do not allow an earlier grouped grant to bypass this final confirmation.

- Re-evaluate the risk when the host, HTTP method, target form, operation, or sensitive fields change. Password and payment fields must retain appropriate protection before data entry.

## Acceptance criteria

- A normal same-site search produces no more than one permission prompt.
- High-risk mutations still require a final pre-dispatch confirmation.
- Sensitive fields remain protected.
- Retries or repeated model attempts do not create duplicate prompts for the same unchanged action.
- The prompt clearly explains the grouped scope and the risk level.
- Tests cover both the low-risk search path and high-risk mutation path.

## Example

Task: `Search Google Scholar for machine learning papers.`

Expected: one concise permission prompt, followed by navigation, query entry, and search submission.

Current behavior: multiple prompts for a single search flow, including a separate form-submission confirmation for the GET search form.

alectimison-maker:
## Summary

- Classify browser-resolved form metadata in pure JavaScript and group only same-site, read-oriented GET searches with non-sensitive fields.
- Add a provisional search-navigation candidate so a normal navigation → query entry → submit flow uses one task-scoped approval.
- Reuse the existing submit probe metadata, mirror the behavior in Chrome and Firefox, and deduplicate unchanged retries per tab and task.
- Keep POST forms, side-effect routes, sensitive fields, `set_field`, JavaScript, WebMCP, and future tasks behind their existing fresh or capability-specific gates.

## Validation

- `npm test`
- 2,389 tests passed
- Security suite: 60/60 checks passed

Manual browser verification is still required before merge.

## Maintainer review request

The classifier is intentionally conservative and does not inspect task wording or page prose. Please confirm that the resolved-URL, registrable-host, side-effect-marker, and field-metadata predicate matches the intended semantics for issue #3060. I can adjust the predicate if the project needs a narrower or broader search surface.

Closes #3060


esokullu:
Pushed fixes addressing the review feedback:

- **Narrow candidate scope**: Bound provisional search candidates to the specific target action base URL and consumed single-use candidate approvals upon promotion so approvals don't cover other searches or forms.
- **Honor existing "always" grants**: Verified persistent capability grants before presenting the grouped search prompt on navigation and typing steps.
- **Inspect hidden fields**: Added hidden control names/IDs to form summaries so forms containing sensitive tokens (`csrf_token`, `api_key`, credentials) are excluded from low-risk grouping.
- **Expanded multi-tenant suffix matching**: Added shared hosting suffixes (e.g. `blogspot`, `wordpress`, `notion.site`, `myshopify.com`) to prevent cross-tenant permission sharing.
- **Candidate lifetime**: Cleared provisional candidates when navigating away from the target site.

Issue #3064, opened 2026-09-19
perf: reduce unnecessary post-click latency in click_ax


alectimison-maker:
## Problem

`click_ax` dispatches the page's synthetic `el.click()` before entering its post-click observation path. When there is no immediate target-state proof, the observer samples after 250 ms and may poll for up to 1,250 ms. For fallback-eligible targets, it can then spend up to another 800 ms in a final settle window before considering a trusted CDP click.

These checks help prevent duplicate submissions and other repeated side effects, but they can make a `click_ax` tool call feel slow even when the page has already responded. The click event itself is dispatched earlier; this issue is about the latency before the agent receives the click result and can continue.

This is separate from [#214](https://github.com/webbrain-one/webbrain/issues/214), which discusses warming the initial accessibility-tree read. This issue focuses on post-dispatch click verification.

## Proposal

- Add stage-level timing for `click_ax`: preflight, synthetic dispatch, post-click observation, trusted-fallback preparation/dispatch, and result return.
- Measure representative immediate, delayed, no-op, navigation, and state-changing clicks before changing the wait policy.
- Evaluate early exits or adaptive observation budgets for cases with reliable, target-specific progress evidence.
- Preserve the current safety behavior when a click may already have caused a side effect or its outcome is unknown.

## Acceptance criteria

- Traces or benchmark output make pre-dispatch, dispatch, observation, fallback, and total tool latency distinguishable.
- A repeatable comparison reports p50/p95 click-to-tool-result latency, overall task success, and duplicate/unsafe dispatches for the current and proposed behavior.
- Strong progress evidence can end observation early without waiting out a fixed timeout.
- Relevant network mutation, download, tab-change, recipient-binding, or unknown-outcome signals continue to suppress an unsafe trusted retry.
- A dispatched click is never blindly replayed because its result was slow or inconclusive.
- No global wait reduction is enabled unless the measurements show lower latency without worse task success or duplicate side effects.

## Out of scope

Removing click verification globally; changing permission or recipient checks; replacing the AX/ref_id path; changing the Firefox synthetic-click behavior; adding a remote judge/model dependency; revisiting initial AX warmup (#214).
```

esokullu:
i believe we can close this because it's not merged into 36.8.0 -- thanks!

PR #3067, opened 2026-09-19
perf: wake click_ax verification from browser events

alectimison-maker:
## Motivation

`click_ax` can wait for its next scheduled sample after the page has already emitted a useful browser signal. This delays returning control while the existing observation and fallback safeguards remain active. Related to #3064.

## Changes

- Wake the Chrome observer on target-tab URL updates, relevant write requests, downloads, and tab creation/removal. Each event only requests a fresh page observation; existing page evidence still decides whether a click is verified or a trusted retry is vetoed.
- Preserve the existing first-sample, polling, maximum-observation, and final-settle budgets. A slow or inconclusive result never triggers a blind replay.
- Add correlated `click_ax` trace notes for preflight, synthetic dispatch and response, observation, trusted-fallback preparation and dispatch, trusted observation, total dispatcher time, and categorical outcomes. The note stores bounded numeric and categorical fields only in both trace privacy tiers.

## Validation

- `node test/run.js`: 2391 passed, 2 failed. Both failures reproduce on the upstream baseline: the changelog ends at 36.7.0 while the package is 36.7.2, and `dist/webbrain-chrome-36.7.2.zip` is absent.
- `npm run test:build-unpacked`: passed (4/4).
- `npm run build:chrome`: passed.
- JavaScript syntax checks and `git diff --check`: passed.
- Paired observer measurements (20 samples per polling/event path; 5 per no-op path):
  - Route proof available at 30 ms: p50/p95 251.1/252.4 ms polling, 31.1/31.3 ms event-woken; proof in 20/20 runs for both.
  - Route proof available at 310 ms: 402.1/403.0 ms polling, 310.8/311.7 ms event-woken; proof in 20/20 runs for both.
  - Relevant POST at 30 ms: 251.3/252.0 ms polling, 31.4/31.5 ms event-woken; 20/20 safety vetoes and no proof in both.
  - No event: 1308.7/1309.6 ms polling, 1309.3/1311.0 ms event-woken; no proof in either path.

These timings come from a controlled observer/snapshot harness, not a live browser or a model-driven task. Overall task-success rates and real-world duplicate-dispatch rates are not measured here; the trace outcomes enable that follow-up. Browser events only trigger earlier sampling; verification, veto, retry, and timeout rules are unchanged.

## Compatibility

The click behavior change is Chrome-only. Firefox synthetic-click behavior is unchanged; its trace privacy projection mirrors the allowlist. No permissions or dependencies were added.


alectimison-maker:
Follow-up commit: `f2240ddf8a9432bcc6a479231a2a5b7bc21964ce`

This follow-up expands the Firefox scope noted in the PR body. Firefox's `click_ax` path did not have Chrome's post-click polling loop; the runtime change targets the existing 80 ms confirmation wait for trusted BiDi checkbox/radio clicks. It reads the requested state immediately, wakes on `input`/`change`, and keeps the 80 ms timeout for cases with no state event. Click timing fields are removed from tool results and recorded through the existing privacy allowlist.

Firefox validation:
- `npm run build:firefox` — passed.
- `npm run test:firefox-bidi` — 24/24 passed.
- `npm run test:firefox-bidi:e2e` — passed in real Firefox; an already-present state returned in 0 ms, and a delayed state event woke observation in 31 ms (80 ms cap).
- `node test/run.js` — 2,391 passed; the same two baseline failures remain: changelog/package version mismatch and missing `dist/webbrain-chrome-36.7.2.zip`.

CI baseline clarification: the previous [WebMCP workflow run](https://github.com/webbrain-one/webbrain/actions/runs/35449322869) stopped at the PDF MIME handler regression on Chrome 152 (`Could not find the tab showing the opaque PDF`), so its WebMCP step was skipped. The workflow now proceeds to WebMCP after setup even when the PDF regression fails. The new [GitHub run](https://github.com/webbrain-one/webbrain/actions/runs/35451119174) passed both the PDF MIME handler regression and the real Chrome WebMCP smoke test. Standalone WebMCP E2E also passed on both the upstream baseline and PR branch with local Chrome 153.

